PR #1, opened 2021-02-25
Module 2 mid assessment complete

Dumin1120:


index.html
```diff
@@ -0,0 +1,30 @@
...
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Dice Roller</title>
+    <script src="index.js"></script>
+</head>
+<body>
+    <h1>Dice Roller</h1>
+    <br>
+    Number of dice:
```
myrasmith, line 13: What are some appropriate HTML tags could you use to encapsulate this text?

index.html
```diff
@@ -0,0 +1,30 @@
...
+    <title>Dice Roller</title>
+    <script src="index.js"></script>
+</head>
+<body>
+    <h1>Dice Roller</h1>
+    <br>
+    Number of dice: 
+    <input type="number" id="input">
+    <br>
+    <button id="button">Roll the dice!</button>
```
myrasmith, line 16: Consider using a form to group related input tags, labels, and submit buttons!

index.html
```diff
@@ -0,0 +1,30 @@
...
+    <br>
+    <div style="border: black solid 3px">
```
myrasmith, line 18: Best practice: separate CSS into its own file.

index.js
```diff
@@ -0,0 +1,35 @@
+document.addEventListener("DOMContentLoaded", () => {
+    let count = 0
+    const button = document.getElementById("button")
+    button.addEventListener("click", () => {
+        const numOfDice = document.getElementById("input").value
+        if(numOfDice){
```
myrasmith, line 6: I like the idea to safeguard this from running if the user didn't input a number. Instead of placing the majority of the click handler inside of this if statement, could you try returning immediately if no user input is provided?

index.js
```diff
@@ -0,0 +1,35 @@
...
+            const diceSum = diceResult.reduce((a, e) => a + e, 0)
+            let diceImage = ""
+            for(let i = 0; i < numOfDice; i++){
+                switch(diceResult[i]){
+                    case 1 : diceImage += "&#9856"; break
+                    case 2 : diceImage += "&#9857"; break
+                    case 3 : diceImage += "&#9858"; break
+                    case 4 : diceImage += "&#9859"; break
+                    case 5 : diceImage += "&#9860"; break
+                    default: diceImage += "&#9861"
```
myrasmith, line 20: Try brainstorming ways to simplify this code to make it more DRY (i.e. Don't Repeat Yourself). You shouldn't have to write 6 nearly identical lines of code.

myrasmith (commented):
Nice work! I left some comments on your HTML and JavaScript for ways you can improve this.

index.js
Dumin1120: updated code in my repo, can't update it here now

PR #2, opened 2021-02-25
Damien Yule

DamienYule:


index.html
```diff
@@ -0,0 +1,24 @@
...
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <script src="index.js"></script>
+    <title>Dice Roll</title>
+</head>
+<body>
+    <H1>Dice Roller</H1>
```
myrasmith, line 11: Why is this tag capitalized?

index.html
```diff
@@ -0,0 +1,24 @@
...
+    Number of Dice: <input type="number" id="number-of-dice"> <br>
```
myrasmith, line 12: What are some appropriate HTML tags you could use to encapsulate this text?

index.html
```diff
@@ -0,0 +1,24 @@
...
+    <button id="btn">Roll the dice!</button>
```
myrasmith, line 13: Consider using a form to group related input, label, and submit buttons.

index.js
```diff
@@ -0,0 +1,54 @@
...
+    const arrayOfdice = [
+        '&#9856',
+        '&#9857',
+        '&#9858',
+        '&#9859',
+        '&#9860',
+        '&#9861',
+    ]
+    button.addEventListener('click', () => {
+        event.preventDefault()
```
myrasmith, line 17: If you need to use an event object, it's best practice to explicitly declare a function parameter for the event, rather than referring to the global `window.event` object.

Is it necessary to prevent the default behavior in this case?

index.js
```diff
@@ -0,0 +1,54 @@
...
+                sum += 2
+            }else if (newArray[i] === '&#9858'){
+                sum += 3 
+            }else if (newArray[i] === '&#9859'){
+                sum += 4
+            }else if (newArray[i] === '&#9860'){
+                sum += 5
+            } else {
+                sum += 6
+            }
```
myrasmith, line 42: This code is pretty repetitive - try brainstorming ways to keep it DRY. You've already got an array with all the dice values in it, maybe you could use that array somehow?

myrasmith (commented):
Nice work! I left some comments on how you can improve this further.

index.html
DamienYule: oops lol.... h1 not H1

index.html
DamienYule: Thanks.. Will do next time

index.js
DamienYule: Yeah, Im still a little shaky on what some of these things are for.

index.html
DamienYule: **<p>** Number of Dice: <input type="number" id="number-of-dice">  **</p>** <br>

DamienYule: <p> </p>

PR #5, opened 2021-02-25
Module-Two-Midmodule-Assessment

PompaDonpa:


index.css
```diff
@@ -0,0 +1,60 @@
+@import url("https://fonts.googleapis.com/css2?family=Press+Start+2P&display=swap");
```
myrasmith, line 1: I love all the work you did to make your assessment look nice!

index.html
```diff
@@ -0,0 +1,28 @@
...
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Dice Roller</title>
+    <link rel="stylesheet" href="./index.css" />
+    <script src="./index.js" defer></script>
+  </head>
+  <body>
+    <h1 class="titles">Dice Roller</h1>
+    <div class="divider" id="goodLooking">
```
myrasmith, line 13: Is there an HTML element that would be more appropriate than a div to group together related inputs, labels, and submit butons?

index.js
```diff
@@ -0,0 +1,60 @@
+// Loading the DOM
+
+document.addEventListener('DOMContentLoaded', () => {
+  // Declaring Global variables, to reuse
```
myrasmith, line 4: These aren't technically global variables since they are declared inside of a callback function.

myrasmith: But I get your meaning - they are all declared in a scope that is visible to your click handler!

index.js
```diff
@@ -0,0 +1,60 @@
...
+
+  // Function that returns a string selected randomly from an array dados[]
+
+  function diceRoll () {
+    // Declaring the vars to be local
+    let fila = ''
+    let faceValue = 0
+
+    // Using a loop to create a string and sum the values depending of
+    // user parameter @param n.value (#input)
```
myrasmith, line 52: The way you've set up your function, n isn't a function parameter - it's a variable in a higher variable scope!
Consider refactoring diceRoll so that it only depends on function parameters instead of variables that live in different scopes.

index.js
```diff
@@ -0,0 +1,60 @@
...
+    for (let i = 1; i <= n.value; i++) {
+      faceValue = Math.floor(Math.random(0) * Math.floor(5))
```
myrasmith, line 54: Are both calls to Math.floor necessary?

index.js
```diff
@@ -0,0 +1,60 @@
...
+    container = `${dice.innerHTML}=${sum}`
+
+    // Restarting the sum to 0 and add 1 to clickCounter
+    sum = 0
+    clickCounter += 1
+  })
+
+  // Function that returns a string selected randomly from an array dados[]
+
+  function diceRoll () {
```
myrasmith, line 46: It looks like this function does two things:
1. returns the string `fila` which stores all the special dice characters
2. Assigns the external variable `sum` to the proper value, but does not return this value.

Could you simplify this function by returning both these values instead of returning one value and setting a variable to the other value?

index.js
```diff
@@ -0,0 +1,60 @@
+// Loading the DOM
+
+document.addEventListener('DOMContentLoaded', () => {
+  // Declaring Global variables, to reuse
+  let dados = ['&#9856;', '&#9857;', '&#9858;', '&#9859;', '&#9860;', '&#9861;']
+  let clickCounter = 0
+  let container = ''
+  let sum = 0
```
myrasmith, line 8: Could you refactor your code so that your click handler doesn't rely on these two variables to function properly?

myrasmith (commented):
I love the work you put into styling your page.

I think your code can be improved by writing functions that depend only on their parameters and on values that they grab directly from the DOM, instead of relying on variables that are declared outside of their own definitions.

index.html
PompaDonpa: Yes, I changed the <div> to <form action=""> same class and id, and also added eve.preventDefault() after eventListener is called.

index.css
PompaDonpa: Thank You so much!

index.js
PompaDonpa: I decided to keep the comments to a minimum, thanks!

PR #6, opened 2021-02-25
Mid Module-2 Assesment

mohsinali678:
Passed All Tests


index.js
```diff
@@ -0,0 +1,97 @@
...
+    dicePara.setAttribute("id", "dice-para");
+    let sumPara = document.createElement("p");
+    sumPara.setAttribute("id", "sum-para");
+    let listItem = document.createElement("li");
+
+
+    if (numberOfDice <= 0) {
+      while (diceParaSumParaDiv.firstChild) {
+        diceParaSumParaDiv.removeChild(diceParaSumParaDiv.firstChild);
+      }
```
jabyess, line 24: curious why you used a `while()` loop here - wouldn't a regular `if()` statement do the same thing? The loop will only ever run once from what I can tell.

index.js
```diff
@@ -0,0 +1,97 @@
...
+  const diceParaSumParaDiv = document.querySelector("#diceParaSumParaDiv");
+  const uList = document.querySelector("#dice-ul");
+  
+
+  diceButtonObject.addEventListener("click", (e) => {
+    const numberOfDice = diceInputObject.value;
+    let dicePara = document.createElement("p");
+    dicePara.setAttribute("id", "dice-para");
+    let sumPara = document.createElement("p");
+    sumPara.setAttribute("id", "sum-para");
```
jabyess, line 17: nice usage of setAttribute! but you can also write these in the HTML directly without having to do any of this.

index.js
```diff
@@ -0,0 +1,97 @@
...
+            sum += dice;
+            break;
+          case dice === 6:
+            str += "&#9861";
+            sum += dice;
+            break;
+          default:
+            console.log("Error in Switch.");
+            console.log(str);
+        }
```
jabyess, line 64: Is there a better way to do this? Any time you have a bunch of conditionals in a row it's often an indicator that you could do re-think your approach.

index.js
```diff
@@ -0,0 +1,97 @@
...
+      listElem.addEventListener('mouseenter', (e)=>{
+      e.target.style.fontSize = '40px';
+    })
+    }
+
+    for(let listElem of uListItem){
+        listElem.addEventListener('mouseleave', (e)=>{
+        e.target.style.fontSize = '16px';
+      })
+    }
```
jabyess, line 87: I love the extra attention to detail! Even though these don't affect the tests in any way they're still fun to implement

index.js
```diff
@@ -0,0 +1,97 @@
...
+            console.log("Error in Switch.");
+            console.log(str);
+        }
+        counter++;
+      }
+      dicePara.innerHTML = str;
+      sumPara.textContent = "Sum = " + sum;
+      diceParaSumParaDiv.appendChild(dicePara);
+      diceParaSumParaDiv.appendChild(sumPara);
+      listItem.innerHTML = str + " = " + sum;
```
jabyess, line 71: Don't forget about using ES6 template strings! It's a personal preference but I think they're easier to read.

```js
listItem.innerHTML = `${str} = ${sum}`
```

index.js
mohsinali678: The loop will run twice, first time for dice para and second time for sum para.
When user enters 0 value it will remove both paras and create another para with unique id
If I remove while loop, selecting 0 value will create two p elements with same id.

index.js
mohsinali678: Yes you are right I could do this, but I wanted to keep div element empty and created elements runtime.

index.js
mohsinali678: Yes there is a better way to use default by setting it to last value which is six.
`
default:
str += "&#9861";
sum += dice;
break;

`

index.js
mohsinali678: I liked that you love it. Thanks

index.js
mohsinali678: That is a precious suggestions.
I will keep that in mind for future.

PR #7, opened 2021-02-25
Finished mid module (Jordan)

JordanAlvarado:


index.html
```diff
@@ -0,0 +1,28 @@
...
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <script src="index.js"></script>
+    <link rel="stylesheet" href="index.css">
+    <title>Document</title>
+  </head>
+  <body>
+    <h1>Dice Roller</h1>
+    number of dice:
```
myrasmith, line 13: What are some appropriate HTML tags you could use to encapsulate this text?

index.html
```diff
@@ -0,0 +1,28 @@
...
+    <input type="number" id="input" placeholder="Enter a Number..." /> <br />
+    <button id="roll-dice" onclick="rollDice()">Roll the dice!</button>
```
myrasmith, line 15: What HTML tag could you use to group together related labels, inputs, and submit buttons?

index.html
```diff
@@ -0,0 +1,28 @@
...
+
+    <div>
+      <p id="invalid"></p>
```
myrasmith, line 18: Nice addition!

index.html
myrasmith, line 15: As you start writing larger, more complex web applications, consider registering click handlers inside your javascript files instead of as inline HTML. This helps keeps different concerns separated: HTML is used for declaring the structure of your content, JavaScript is used for creating and configuring responses to user events.

index.js
```diff
@@ -0,0 +1,45 @@
...
+let oldRolls = []
+
+function rollDice(){
+    let invalid = document.getElementById('invalid')
+    let diceTotal = document.getElementById('dice-para')
+    let sum = document.getElementById('sum-para')
+    let ul = document.querySelector('#roll-history')
+    let input = Number(document.getElementById('input').value)
+
+    if(input > 0){
```
myrasmith, line 20: I love the idea of safeguarding your core logic so that it only runs if the user input is valid. Instead of placing all of the core logic of you click handler inside of this if statement, could you check for invalid user input first, and return early from the function?

index.js
```diff
@@ -0,0 +1,45 @@
...
+            diceTotal.innerHTML += diceImg[diceRolled]
+            sumDice += diceRolled
+        }
+        inputHistory++
+        sum.innerText = "Sum = " + sumDice
+        oldRolls[inputHistory] = diceTotal.innerHTML + " = " + sumDice
+
+        if(inputHistory > 1){
+            let newList = document.createElement("li")
+            newList.textContent = oldRolls[inputHistory -1]
```
myrasmith, line 36: keeping track of oldRolls and inputHistory strikes me as more work than necessary. Could you refactor your code to achieve the same functionality without relying on these variables?

myrasmith (commented):
Nice work!

index.html
JordanAlvarado: I would have to use <form> to group <input> and <button> together

PR #8, opened 2021-02-25
Completed Mid-Module 2 Assessment

bongster91:


index.css
```diff
@@ -0,0 +1,27 @@
...
+
+#sum-para {
+    margin-top: 5px;
+    margin-bottom: 5px;
+    font-size: 30px;
+}
+
+ul {
+    font-size: 30px;
+}
```
jabyess, line 27: I appreciate that you spent some time on CSS even though it doesn't affect the tests passing at all!

index.js
```diff
@@ -0,0 +1,63 @@
+const one = `&#9856`
+const two = `&#9857`
+const three = `&#9858`
+const four = `&#9859`
+const five = `&#9860`
+const six = `&#9861`
```
jabyess, line 6: Instead of having six different variables, what kind of data structure could you use to store a collection of related things?

index.js
```diff
@@ -0,0 +1,63 @@
...
+        let numArr = []
+        let sum 
+
+        function sumDie () {
+            count += 1
+            for (let i = 0; i < input; i += 1) {
+                numArr.push(Math.floor(Math.random() * 6 + 1))
+            }
+
+            sum = numArr.reduce((a, b) => a + b)
```
jabyess, line 26: Love this usage of reduce!

index.js
```diff
@@ -0,0 +1,63 @@
...
+                   case 3: img += three;
+                   break
+                   case 4: img += four;
+                   break
+                   case 5: img += five;
+                   break
+                   case 6: img += six;
+                   break
+               }
+           })
```
jabyess, line 49: Is there a better way to do this? Anytime i have a bunch of if/else statements in a row, its a sign that I need to probably re-think my pattern. This might be more obvious once you collect all six of these variables into one thing (see my comment above).

index.js
```diff
@@ -0,0 +1,63 @@
...
+           dieOutput.innerHTML = img
+           newArr.push({dieImg: img, dieSum: sum})
```
jabyess, line 51: I love arrays of objects!!!!

index.js
```diff
@@ -0,0 +1,63 @@
...
+        }
+        displayDie()
+     
+        if (count > 1) {
+            let list = document.querySelector('ul')
+            let li = document.createElement('li')
+            li.innerHTML = `${newArr[0].dieImg} = ${newArr[0].dieSum}`
+            list.appendChild(li)
+            newArr.shift()
```
jabyess, line 60: I haven't seen anyone else use shift here! this is a great solution.

index.js
bongster91: An object! Duh! (face-palm)

index.js
bongster91: I was so proud I remembered to use reduce

index.js
bongster91: Part of the object for the dice images above? Like one attribute is the image, one is the numerical number and one as the written out number?

PR #9, opened 2021-02-25
This is as far as I can get right now

jasleenv:


index.html
```diff
@@ -0,0 +1,57 @@
+<!DOCTYPE html>
+
+<html>
+<head>
+    <h1> Dice Roller </h1>
+    <meta charset="UTF-8"
+</head>
+<body>
+   <p>Number of dice:  <input type="number" id="numberOfDice"> </p>
+    <button type="submit" onclick="rollDice()" id="dice-para" >Roll the dice! </button> </p>
```
myrasmith, line 10: What is the purpose of this button element?
What is the intended purpose of the HTML element with id="dice-para"?

Consider creating a separate HTML element whose dedicated purpose is to store the dice values that were rolled on the most recent roll.

index.html
```diff
@@ -0,0 +1,57 @@
...
+<h2>History</h2>
+    <ul id="history"></ul>
+    <p id ="sum-para"></p>
+
+    <script>
+        function rollDice() {
+            const dice = ['&#9856 = ⚀','&#9857 = ⚁','&#9858 = ⚂','&#9859 = ⚃','&#9860 = ⚄','&#9861 = ⚅']
```
myrasmith, line 17: I like that you are using an array to store the dice!

The HTML code '&#9856' and the emoji  '⚀' are two ways of representing the same character, so you don't have to use both forms.

index.html
```diff
@@ -0,0 +1,57 @@
...
+       return dice
```
myrasmith, line 18: Why are you returning the dice here?
What happens after a function returns? Does it continue executing the rest of the code inside the function or does it immediately stop?

index.html
```diff
@@ -0,0 +1,57 @@
...
+            let amountOfDice = document.getElementById("numberOfDice").value
+            let dicePics = Math.floor(Math.random()* dice.length)
+            return dicePics
+          
+            let sumoOfDice = dicePics.value
```
myrasmith, line 23: What type of data is dicePics?
Does dicePics have a .value property?

index.html
```diff
@@ -0,0 +1,57 @@
...
+    <meta charset="UTF-8"
+</head>
+<body>
+   <p>Number of dice:  <input type="number" id="numberOfDice"> </p>
+    <button type="submit" onclick="rollDice()" id="dice-para" >Roll the dice! </button> </p>
+<h2>History</h2>
+    <ul id="history"></ul>
+    <p id ="sum-para"></p>
+
+    <script>
```
myrasmith, line 15: Consider writing your javascript in a separate file! As you start writing more complex web apps, you'll be writing hundreds or thousands of lines of javascript code so it gets messy to keep it inside your HTML files.

myrasmith (commented):
There are some definite positive takeaways here:
1. You are grabbing user input from the DOM correctly
2. You are creating a new DOM element and appending it to the DOM
3. You are registering the click handler (rollDice) correctly using the 'onclick' attribute

But there are also some things you need to work on:
1. Control flow. For example, using 'return' will stop a function in its tracks and it won't continue executing after it hits a return statement.
2. Recognizing the data type of a variable. For example, dicePics is a number, not a DOM element, so it doesn't have a `.value` property.
3. Applying the JavaScript fundamentals from Module 1 to implement application logic. For example, how could you use a for loop to create a string of random dice characters?

A few general notes: 
- if you struggle interpreting the README instructions, please reach out with questions! We are never trying to ask trick questions or give bad instructions. If the instructions aren't clear, it's part of my job to help make them more clear.
- when you're debugging and want to see the value stored in some variable, use a console.log - not a return statement! You can see your console logs in the chrome dev tools.

PR #10, opened 2021-02-25
Cassidy Beni

cassidybeni:


index.js
```diff
@@ -0,0 +1,40 @@
...
+    btn.addEventListener('click', () => {
+
+        let input = Number(document.querySelector('#number-of-dice').value)
+        let dicePara = document.querySelector('#dice-para')
+        let sumPara = document.querySelector('#sum-para')
+
+        let str = ""
+        for(let i = 0; i < input; i++){
+            str += dice[Math.floor(Math.random() * dice.length)]
+            dicePara.innerText = str
```
jabyess, line 14: I see that you figured out you have to use .innerText instead of .textContent. Nice!

index.js
```diff
@@ -0,0 +1,40 @@
...
+            } else if(str[i].includes(dice[2])){
+                sum += 3
+            } else if(str[i].includes(dice[3])){
+                sum += 4
+            } else if(str[i].includes(dice[4])){
+                sum += 5
+            } else if(str[i].includes(dice[5])){
+                sum += 6
+            }
+        }
```
jabyess, line 32: Is there a better way to do this other than with a bunch of if/else statements? Any time you find yourself writing a whole lot of these, try to see if there's a different pattern you can follow.

index.js
```diff
@@ -0,0 +1,40 @@
...
+        
+        let ul = document.querySelector('#history-list')
+        let li = document.createElement('li')
+        li.innerText = `${dicePara.innerText} = ${sum}`
```
jabyess, line 36: Love that you combined a property here and a plain string variable, all inside an es6 template string. I dunno if I've seen anyone else do that.

PR #11, opened 2021-02-25
First Commit

Dequisa:
Last Break before final submit

### DIFF
--- a/style.js
+++ b/style.js
@@ -1,0 +1,78 @@
+console.log("This is Your Dice Game")
+
+document.addEventListener("DOMContentLoaded",()=>{
+    const dice = ['&#9856;','&#9857;','&#9858;','&#9859;','&#9860;','&#9861;']
+
+
+    let form = document.querySelector("form")
+    form.addEventListener("submit", (event)=>{
+        event.preventDefault()
+
+let rollButton = document.querySelector('#start-game')
+        let num = 0;
+        let array = []
+        let input = document.querySelector('input')
+        let inputValue = input.value;
+        
+    rollButton.addEventListener("click",(e)=>{
+        // para.innerHTML= ""
+        // diceDiv.appendChild(para)
+        for(let i = 1;i<=inputValue;i++){ 
+        e.preventDefault()
+        let para = document.createElement("p")
+        let diceDiv = document.getElementById("dice-para")
+        let history = document.querySelector("ul")
+        let sum = document.getElementById("total-sum")
+        let createLi = document.createElement("li")
+        
+        
+        
+        
+        
+       
+      let random = Math.floor(Math.random()* 6) 
+    
+        if(random ===1){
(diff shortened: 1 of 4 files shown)


style.js
```diff
@@ -0,0 +1,90 @@
...
+                totSum.push(dice[5])  
+                num +=6 
+                array.push(dice[5])
+                }
+        // sum.textContent = `Sum = ${num}`;
+        
+            }
+            sum.textContent = `Sum = ${num}`;
+            para.innerHTML=" "
+            totSum.map((e)=>{para.innerHTML += e })
```
myrasmith, line 67: It looks to me like you are using this line for it's side effects (adding onto para.innerHTML), not for the return value of the `.map` method. In that case, try using `.forEach` instead!

style.js
```diff
@@ -0,0 +1,90 @@
...
+            diceDiv.appendChild(para)
+            para.parentNode.replaceChild(para, para)
```
myrasmith, line 69: Why are you replacing para with itself? Is it necessary to replace a DOM element if you have altered its innerHTML already?

style.js
```diff
@@ -0,0 +1,90 @@
...
+
+
+    let form = document.querySelector("form")
+    form.addEventListener("submit", (event)=>{
+        event.preventDefault()
+
+let rollButton = document.querySelector('#start-game')
+        let num = 0;
+        let array = []
+        let totSum = [];
```
myrasmith, line 14: Do you need both of these arrays? Could you use a single array to achieve the same functionality?

style.js
```diff
@@ -0,0 +1,90 @@
...
+        
+    rollButton.addEventListener("click",(e)=>{
```
myrasmith, line 16: If you registered this event listener outside of the form submittion listener function, how would your application behave differently?

style.js
```diff
@@ -0,0 +1,90 @@
...
+            }
+            sum.textContent = `Sum = ${num}`;
+            para.innerHTML=" "
+            totSum.map((e)=>{para.innerHTML += e })
+            diceDiv.appendChild(para)
+            para.parentNode.replaceChild(para, para)
+            // diceDiv.appendChild(para)
+            console.log(diceDiv)
+            console.log(para.innerHTML)
+             array.map((e)=>{createLi.innerHTML += e })
```
myrasmith, line 73: .forEach is more appropriate here, as well! Could you combine this step with the totSum.map step somehow?

myrasmith (commented):
There are a bunch of great things happening in your code!
- grabbing input from the DOM well
- Creating a string of random dice and calculating their sum
- Creating and appending elements to the DOM well

There are also some areas you can improve:
- Make your code DRY (Don't Repeat Yourself)
- Don't create multiple variables that store the same data
- When/where you call the `.addEventListener` method matters

PR #12, opened 2021-02-25
Module 2 TA

tiffanyarnold:


index.js
```diff
@@ -0,0 +1,42 @@
...
+    roll.addEventListener("click", (event) => {
+        click++;
+        const list = [
+            { dice: "&#9856", value: 1 },
+            { dice: "&#9857", value: 2 },
+            { dice: "&#9858", value: 3 },
+            { dice: "&#9859", value: 4 },
+            { dice: "&#9860", value: 5 },
+            { dice: "&#9861", value: 6 },
+        ];
```
jabyess, line 14: Love me an array of objects! This makes your loop below much simpler to implement.

index.js
```diff
@@ -0,0 +1,42 @@
...
+        const diceNumber = Number(document.querySelector("#num-of-dice").value);
+        const result = document.querySelector("#dice-para");
+        const diceAdd = document.querySelector("#sum-para");
+        const history = document.querySelector("div > ul");
```
jabyess, line 18: You could just do `querySelector('ul')` here because there is only one `ul` on the whole page. But this is actually safer and a better practice!

index.js
```diff
@@ -0,0 +1,42 @@
...
+        diceAdd.textContent = "Sum = " + currentAdd;
+        if (click === 0) {
+            newLi = document.createElement("li").style.visibility = "hidden";
+            newLi.innerHTML = `${rollTwo}=${historyAdd}`;
+            history.appendChild(newLi);
+        } else if (click > 1) {
+            newLi = document.createElement("li");
+            newLi.innerHTML = `${rollTwo}=${historyAdd}`;
+            history.appendChild(newLi);
+        }
```
jabyess, line 38: You have an almost identical solution to Hady's, so I'm just gonna repeat what I said there:

This is a great solution, though you should know that even if an element has visibility: hidden, it can still be selected by cypress or querySelector. the reason this passes is because cypress only checks the history after clicking the button several times.

PR #14, opened 2021-02-25
finished

HadyM:


index.js
```diff
@@ -0,0 +1,50 @@
...
+      { dice: "&#9860", value: 5 },
+      { dice: "&#9861", value: 6 },
+    ];
+
+    const numOfDice = Number(document.querySelector("#num-of-dice").value);
+    const result = document.querySelector("#dice-para");
+    const sumOfDice = document.querySelector("#sum-para");
+    const history = document.querySelector("div > ul");
+
+    event.preventDefault();
```
jabyess, line 23: Generally you want to put event.preventDefault() as the __very first__ thing in any event listener callback. In this instance it doesn't matter, but it's a good habit to get into.

index.js
```diff
@@ -0,0 +1,50 @@
...
+  button.addEventListener("click", (event) => {
+    numClicks++;
+    const diceList = [
+      { dice: "&#9856", value: 1 },
+      { dice: "&#9857", value: 2 },
+      { dice: "&#9858", value: 3 },
+      { dice: "&#9859", value: 4 },
+      { dice: "&#9860", value: 5 },
+      { dice: "&#9861", value: 6 },
+    ];
```
jabyess, line 16: Love how you structured this! Arrays of objects are your friend.

index.js
```diff
@@ -0,0 +1,50 @@
...
+
+    event.preventDefault();
+    let currentRoll = [];
+    let currentSum = 0;
+
+    for (let i = 0; i < numOfDice; i++) {
+      let randomNum = Math.floor(Math.random() * Math.floor(diceList.length - 1));
+      let dice = diceList[randomNum]["dice"];
+      currentRoll.push(dice);
+      currentSum += diceList[randomNum]["value"];
```
jabyess, line 31: I don't have any real feedback here other than: awesome work, this is a super clean solution.

index.js
```diff
@@ -0,0 +1,50 @@
...
+
+    if (numClicks === 0) {
+      newList = document.createElement("li").style.visibility = "hidden";
+      newList.innerHTML = `${historyRoll} = ${historySum}`;
+      history.appendChild(newList);
+    } else if (numClicks > 1) {
+      newList = document.createElement("li");
+      newList.innerHTML = `${historyRoll} = ${historySum}`;
+      history.appendChild(newList);
+    }
```
jabyess, line 45: This is a great solution, though you should know that even if an element has `visibility: hidden`, it can still be selected by cypress or querySelector. the reason this passes is because cypress only checks the history after clicking the button several times.

HadyM: Hey Jimmy thank you for the feedback!!! I have a question about cypress you said that it checks after clicking the button several times. I'm confused like should I of went with 'display: none' instead of 'visibility: hidden'. I think I misread the question in the test I thought it wanted if no input is put and the button is pressed to not display anything but still keep the list...

index.js
HadyM: Got it thank you I will get myself in the habit of doing this.

PR #15, opened 2021-02-25
module 2 mid-assessment

Kelvinz6820:


index.html
```diff
@@ -0,0 +1,26 @@
...
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Dice Rolling</title>
+    <script src="diceRolling.js" defer></script>
+    <link rel="stylesheet" href="diceRolling.css">
+</head>
+<body>
+    <h1>Dice Roller</h1><br>
+    Number of dice: <input type="number" id = "dice-input" placeholder = "Enter a number">
```
myrasmith, line 13: What are some appropriate HTML elements you could use to encapsulate the 'Number of dice:' text?

index.html
```diff
@@ -0,0 +1,26 @@
...
+    <button type = "submit" id = "dice-button" onclick="diceRoll()">Roll the dice!</button>
+    <div class ="first-box">
```
myrasmith, line 15: Consider using a form tag to group related inputs, labels, and submit buttons!

index.html
```diff
@@ -0,0 +1,26 @@
...
+        <p id="invalid"></p>
```
myrasmith, line 16: Nice addition!

index.html
```diff
@@ -0,0 +1,26 @@
...
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Dice Rolling</title>
+    <script src="diceRolling.js" defer></script>
+    <link rel="stylesheet" href="diceRolling.css">
+</head>
+<body>
+    <h1>Dice Roller</h1><br>
+    Number of dice: <input type="number" id = "dice-input" placeholder = "Enter a number">
+    <button type = "submit" id = "dice-button" onclick="diceRoll()">Roll the dice!</button>
```
myrasmith, line 14: I'd recommend refactoring so that your diceRoll function is registered as an event listener in your javascript file, instead of inline here. As your web applications grow in size and complexity, separation of concerns becomes more important! HTML is good for declaring the structure of your content, but JavaScript is more appropriate for creating and configuring how your app responds to user events.

diceRolling.js
```diff
@@ -0,0 +1,41 @@
...
+let historyCounter = 0
+let recordedHistory = []
+
+function diceRoll () {
+    let invalid = document.querySelector("#invalid")
+    let diceInput = Number(document.querySelector("#dice-input").value)
+    let diceDisplay = document.querySelector("#dice-para")
+    let displaySum = document.querySelector("#sum-para")
+    let diceHistory = document.querySelector("#dice-history")
+    if (diceInput > 0) {
```
myrasmith, line 11: I love your idea to safeguard your code so that it only runs if the user input is valid. Instead of placing most of the core logic of your click handler inside one big if statement, could you check for invalid input first and return early from the function?

diceRolling.js
```diff
@@ -0,0 +1,41 @@
+let diceOptions = ['&#9856', '&#9857', '&#9858', '&#9859', '&#9860', '&#9861']
+let historyCounter = 0
+let recordedHistory = []
```
myrasmith, line 3: Could you refactor your code without using these variables?

myrasmith (commented):
Nice work!

PR #16, opened 2021-02-25
Initial commit

kathypurry:


diceRoll.js
```diff
@@ -0,0 +1,110 @@
...
+                sumOutput +=3
+            } else if (arr[i] === "&#9859") {
+                sumOutput += 4
+            } else if (arr[i] === "&#9860") {
+                sumOutput += 5
+            } else if (arr[i] === "&#9861") {
+                sumOutput += 6
+            }
+            
+        }
```
jabyess, line 42: Is there a better way you could do this other than with a bunch of if/else statements? Since you have everything already in an array, what if you used the array index to add to the sum?

diceRoll.js
```diff
@@ -0,0 +1,110 @@
...
+        let newRoll = document.createElement('li')
+        newRoll.innerHTML = `${sumOutput} ${arr.join('')}`
+        ul.appendChild(newRoll)
+
+        sum.innerHTML = `Sum = ${sumOutput}`
+        diceOutput.innerHTML = arr.join('')
```
jabyess, line 48: Glad you figured out the issues with getting the dice to show! I think .innerHTML is the secret.

diceRoll.js
```diff
@@ -0,0 +1,110 @@
...
+//         for (let i = 0, i > 0; i++) {
+            
+//         }
+
+
+//     })
+
+//     const diceRolled = document.createElement('li)')
+//     diceRolled.textContent = input
+//     diceHistory.appendChild(diceRolled)
```
jabyess, line 110: If you are done and especially once you've got all the tests passing, I would delete all this code! It's a good habit to never commit commented-out code.

PR #18, opened 2021-02-25
Commit One

imrannazir20:


index.js
```diff
@@ -0,0 +1,62 @@
...
+rollBtn.addEventListener('click', rollClicked)
+function rollClicked(){
+console.log('Roll the dice button pressed')
+let number =getNumberOfDice()
+if (number === 0){
+    sumPara.innerHTML="Please insert a value"
+    return;
+}
+createDices(number)
+}
```
jabyess, line 24: I love how you created separate functions and passed in arguments! That's a great way to ensure code reuse, and in general is a good pattern to follow.

Don't forget to use your code formatter here!

index.js
```diff
@@ -0,0 +1,62 @@
...
+function addToHistoryList(result, sum){
+let newLi=document.createElement('li')
+newLi.innerHTML= result+ "=" +sum
+let dicePara= document.querySelector('#dice-para')
+dicePara.appendChild(newLi)
+}
+function getNumberOfDice(){
+    return Number(inputBox.value)
+}
```
jabyess, line 60: If you notice your overall pattern here, you are writing functions like this:

```js
function doThingOne() {
  // do stuff
  doThingTwo()
}

function doThingTwo() {
  // do more stuff
  doThingThree()
}

function doThingThree() {
  // do even more stuff
  doThingFour()
}
```

This pattern works, but it is pretty constricting. You're limited to how much you can change without causing cascading changes down the line.

I might suggest rearranging your code like so:

```js
function doAllThings() {
  let thingOneResult = doThingOne()
  let thingTwoResult = doThingTwo()
  let thingThreeResult = doThingThree()
}

function doThingOne() {
  // do stuff
  return result
}

function doThingTwo() {
  // do stuff
  return result
}

function doThingThree() {
  // do stuff
  return result
}
```

This pattern lets you move things around more - you can rearrange the order if you need, and can more easily change how much stuff one function does without affecting the others.

index.js
```diff
@@ -0,0 +1,62 @@
...
+        }
+        saveToHistory(dices)
+    }
+}
+function saveToHistory(dices){
+    let sum =0;
+    let result =''
+    for(i =0; i<dices.length; i++){
+        let currentDice= dices[i]
+        result =result+''+ diceLetters[currentDice-1]
```
jabyess, line 44: You should be able to simplify this just by doing

```js
result += diceLetters[currentDice - 1]
```

index.js
```diff
@@ -0,0 +1,62 @@
...
+    '&#9859',
+    '&#9860',
+    '&#9861'
+    ]
+const rollBtn = document.querySelector('#roll-btn')
+const inputBox =document.querySelector('#dice-number-input')
+//const history = document.querySelector('#dice-para')
+const sumPara = document.querySelector("#sum-para")
+let rollNumber = 0
+rollBtn.addEventListener('click', rollClicked)
```
jabyess, line 15: You're the first person I've seen use the eventListener callback in this way! most people passed in an inline arrow function.

index.js
imrannazir20: Thank you @Jimmy. I like this idea!

PR #19, opened 2021-02-25
Mid-Assessment 

jennaqian:


index.js
```diff
@@ -0,0 +1,17 @@
...
+    let diceInput = document.getElementById("dice-input")
+    let input = diceInput.value
+
+    let possibleOutcomes = ["&#9856 = ⚀", "&#9857 = ⚁", "&#9858 = ⚂", "&#9859 = ⚃", "&#9860 = ⚄", "&#9861 = ⚅"]
+    let sum = 0
+
+    for(let i = 1; i <= input; i++) {
+        let results = (Math.round(Math.random * 6))
+        sum = sum + results
+    }
```
myrasmith, line 11: I think you've got a nice strategy here. There are some problems with the implementation, though!
- how could you call the Math.random function to return a random number between 0 and 1?
- could you use this loop to create a string of random dice emojis as well?

index.js
```diff
@@ -0,0 +1,17 @@
...
+  
+    let uList = document.querySelector("dice-history")
+    let liHistory = document.createElement('li')
+    ul.appendChild(liHistory)
```
myrasmith, line 15: What content should be in this list item?

myrasmith (commented):
Here are some positives:
- you are grabbing the input from the DOM well!
- you are using a loop to create the right amount of random dice rolls
- you are creating a DOM element and appending it to the DOM

Here are some things you can work on:
- What is the difference between a function's name and calling a function?
- Once you've created the sum, what should you do with it?
- How do you add new content into a DOM element?

PR #21, opened 2021-02-25
suger bears

TamzeedRahman:


index.html
```diff
@@ -0,0 +1,27 @@
...
+        <title>Document</title>
+        
+    
+    </head>
+    <body>
+    
+        <h1>Dice Roller</h1>
+       
+            
+           Number of dice: <input id="number-of-dice" type="number">
```
myrasmith, line 16: What are some appropriate HTML tags you could use to encapsulate the 'Numer of dice:' text?

index.js
```diff
@@ -0,0 +1,29 @@
...
+    '&#9860':5, 
+    '&#9861':6
+        }
+
+    document.addEventListener('DOMContentLoaded', () => {
+        let input = Number(document.querySelector('#number-of-dice').value)
+        let rollTheDice = document.querySelector('#roll-the-dice')
+        let diceStory = document.querySelector('#dice-para')
+        rollTheDice.addEventListener('click', () => {
+            roll()
```
myrasmith, line 15: Since your callback just invokes another function, is there a way you could simplify this?

index.html
```diff
@@ -0,0 +1,27 @@
+ <!DOCTYPE html>
+    <html lang="en">
+    <head>
+        <meta charset="UTF-8">
+        <meta http-equiv="X-UA-Compatible" content="IE=edge">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <title>Document</title>
+        
+    
+    </head>
```
myrasmith, line 10: Add a script tag to load your javascript file!

index.js
```diff
@@ -0,0 +1,29 @@
...
+    '&#9856':1,
+    '&#9857':2, 
+    '&#9858':3, 
+    '&#9859':4, 
+    '&#9860':5, 
+    '&#9861':6
+        }
+
+    document.addEventListener('DOMContentLoaded', () => {
+        let input = Number(document.querySelector('#number-of-dice').value)
```
myrasmith, line 11: How could you make sure this input number gets updated each time a user clicks the button?

index.js
```diff
@@ -0,0 +1,29 @@
...
+        let diceStory = document.querySelector('#dice-para')
+        rollTheDice.addEventListener('click', () => {
+            roll()
+        })
+        function roll(){
+           let diceDsply = ''
+            for (let i = 0; i < input; i++) {
+
+                let arrKeys = Object.keys(dice) //[array of dice]
+                 diceDsply = arrKeys[floor(Math.random() * 6)] //
```
myrasmith, line 22: What is floor?

myrasmith: Also, instead of reasigning diceDsply, how could you add a single new character to the end of it with each iteration?

myrasmith (commented):
I think you've got a nice start on a strategy for creating a random dice roll each time. Here are a couple questions that might help getting past some roadblocks:

- How can you make sure your code runs when the browser loads your HTML?
- How can you get the value input a user enters **each time the button is clicked**?

index.js
TamzeedRahman: alittle  confused by what you mean

index.js
TamzeedRahman: += ?

PR #22, opened 2021-02-25
Mid-Module Assessment Update

sbweins:


index.js
```diff
@@ -0,0 +1,42 @@
...
+      } else if (arr[i] === diceArr[2]) {
+        sum += 3;
+      } else if (arr[i] === diceArr[3]) {
+        sum += 4;
+      } else if (arr[i] === diceArr[4]) {
+        sum += 5;
+      } else if (arr[i] === diceArr[5]) {
+        sum += 6;
+      }
+    }
```
jabyess, line 33: This code works! Is there a cleaner way of doing this though? Often when I have a bunch of if/else statements in a row it's a sign that I could structure my code differently to avoid having to type all this out.

jabyess: Maybe you could use the index of diceArr?

index.js
```diff
@@ -0,0 +1,42 @@
...
+
+    const diceLi = document.createElement("li");
+    diceLi.innerHTML = `${arr.join("")} = ${sum}`;
```
jabyess, line 36: Good usage of `join` here! I see lots of mistakes where people tried to append the whole array to the string, but that doesn't work. Clearly you already figured that out.

PR #23, opened 2021-02-25
Module 2 Assessment 

Tpichardo:


index.html
```diff
@@ -0,0 +1,20 @@
...
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Document</title>
+    <script src="./index.js"></script>
+</head>
+<body>
+    <h1 id="first">Dice Roller</h1>
+    Number of dice:<input type="number">
```
myrasmith, line 12: What HTML tags could you use to encapsulate the 'Number of dice:' text?

index.html
```diff
@@ -0,0 +1,20 @@
...
+    <button type="button">Roll the dice!</button>
```
myrasmith, line 13: Consider using a form to group related inputs, labels, and submit buttons!

index.js
```diff
@@ -0,0 +1,43 @@
...
+        { dice: "&#9857", value: 2 },
+        { dice: "&#9858", value: 3 },
+        { dice: "&#9859", value: 4 },
+        { dice: "&#9860", value: 5 },
+        { dice: "&#9861", value: 6 },
+    ];
+    
+    btn.addEventListener("click", () => {
+        random();
+    });
```
myrasmith, line 19: Since this callback just invokes another function, could you simplify this code?

index.js
```diff
@@ -0,0 +1,43 @@
+document.addEventListener("DOMContentLoaded", () => {
+    const btn = document.querySelector("button");
+    const numInput = document.querySelector("input");
+    const dicePara = document.querySelector("#dice-para");
+    const sumPara = document.querySelector("#sum-para");
+    const ul = document.createElement("ul")
+    document.body.appendChild(ul)
```
myrasmith, line 7: Why did you choose to create the ul here instead of putting it in your HTML?

index.js
```diff
@@ -0,0 +1,43 @@
...
+        dicePara.innerHTML = ""
+        for (let i = 0; i < numInput.value; i++) {
+            let randomNumber = Math.floor(Math.random() * arr.length)
+            dicePara.innerHTML += arr[randomNumber].dice
+            sum += arr[randomNumber].value
+        };
+        sumPara.textContent = `Sum = ${sum}`
+        
+    numOfClicks++
+        if(numOfClicks > 0){
```
myrasmith, line 34: Will this ever be false?

myrasmith (commented):
Nice work!

index.js
Tpichardo: I chose to create the ul in my JS file because the instructions did not specify where to create it, and I thought you guys wanted us to practice creating it in JS.

index.js
Tpichardo: This will only be false if the user has not clicked. I actually updated the code on this. If you can, I would love your feedback on it!

index.js
Tpichardo: Yes, would making random() my callback be better?

index.html
Tpichardo: Thanks Myra! Will definitely keep this in mind.

index.html
Tpichardo: I could encapsulate it in a label! 😅

PR #24, opened 2021-02-25
Mid Module Assessment

esayh:


dice.js
```diff
@@ -0,0 +1,40 @@
+const diceArr = [
+	{ dice: '&#9856', value: 1 },
+	{ dice: '&#9857', value: 2 },
+	{ dice: '&#9858', value: 3 },
+	{ dice: '&#9859', value: 4 },
+	{ dice: '&#9860', value: 5 },
+	{ dice: '&#9861', value: 6 },
+]
```
jabyess, line 8: Clever usage of mixing the arrays and objects! This makes your code below much simpler.

dice.js
```diff
@@ -0,0 +1,40 @@
...
+document.addEventListener('DOMContentLoaded', () => {
+	const input = document.querySelector('input')
+	const history = document.querySelector('#history')
+	const btn = document.querySelector('button')
+	const para = document.querySelector('#dice-para')
+	const sum = document.querySelector('#sum-para')
+	const ul = document.createElement('ul')
+	document.body.appendChild(ul)
```
jabyess, line 16: You can do this in the HTML by manually adding a `ul`, but this totally works too. Just wondering why you did it this way

dice.js
```diff
@@ -0,0 +1,40 @@
...
+		para.innerHTML = ''
+		for (let i = 0; i < input.value; i++) {
+			let randDice = Math.floor(Math.random() * diceArr.length)
+			para.innerHTML += diceArr[randDice].dice
+			add += diceArr[randDice].value
+		}
+
+		sum.textContent = `Sum = ${add}`
+
+		clicks++
```
jabyess, line 33: This counter object is a really clear and simple way of handling this. Combined with the if statement below, its a good solution!

dice.js
```diff
@@ -0,0 +1,40 @@
...
+		getDice()
+	})
+	let clicks = 0
+	function getDice() {
+		let add = 0
+		para.innerHTML = ''
+		for (let i = 0; i < input.value; i++) {
+			let randDice = Math.floor(Math.random() * diceArr.length)
+			para.innerHTML += diceArr[randDice].dice
+			add += diceArr[randDice].value
```
jabyess, line 28: These two lines reduce so much complexity of code. Because you set up the array above in the right way, all you have to do is get the properties. Nice work!

dice.js
```diff
@@ -0,0 +1,40 @@
...
+		}
+
+		sum.textContent = `Sum = ${add}`
+
+		clicks++
+		if (clicks > 0) {
+			const li = document.createElement('li')
+			ul.appendChild(li)
+			li.textContent = `${para.innerHTML} = ${add}`
```
jabyess, line 37: Love me some es6 template strings

dice.js
esayh: yea I don't know why I didn't think about that during the test. I guess I was having the same thought process as the emoji story for some reason.

PR #25, opened 2021-02-25
HARD

rayfinn55:


index.html
```diff
@@ -0,0 +1,23 @@
...
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Document</title>
+    <link rel="stylesheet" href="index.css">
+    <script src="index.js"></script>
+</head>
+<body>
+    <h1>Dice Roller</h1>
+    Number of Dice: <input id="dice-input" type="number">
```
myrasmith, line 12: What HTML element could you use to encapsulate the 'Number of Dice:' text?

index.html
```diff
@@ -0,0 +1,23 @@
...
+    <button type="submit" id="dice-roller" onclick="diceRoller()">Roll The Dice!</button>
```
myrasmith, line 13: Consider using a form to group related inputs, labels, and submit buttons!

myrasmith, line 13: How could you refactor your code to register your click handler function in your javascript, instead of inline in your HTML?

index.js
```diff
@@ -0,0 +1,25 @@
...
+    { dice: "&#9859", value: 4 },
+    { dice: "&#9860", value: 5 },
+    { dice: "&#9861", value: 6 },
+  ];
+
+function diceRoller() {
+    const roll = document.getElementById('dice-input')
+    const rollInput = roll.value 
+    output = document.getElementById('sum-para')
+    for(i=1; i< rollInput.length; i++) {
```
myrasmith, line 14: JavaScript best practice: declare the variable i using the let keyword!

index.js
```diff
@@ -0,0 +1,25 @@
...
+        roll += sum
```
myrasmith, line 15: where is the `sum` variable defined? Is `roll` a number/string that you can use the += to change?

index.js
```diff
@@ -0,0 +1,25 @@
...
+        const historyList = document.createElement('li')
+        historyList.textContent += rollInput
+        output.appendChild(historyList)
```
myrasmith, line 18: Why is this code inside your for loop?
What content do you want this list item to contain?
What HTML element do you want to append this list item to?

index.html
rayfinn55: <p> tag  🤦🏾‍♂️

index.html
rayfinn55: so i thought you guys said that events weren't on the test so i don't know why i figured that meant  DOM content loaded and event listeners were also not on the test.

index.js
rayfinn55: 🤦🏾‍♂️

PR #26, opened 2021-02-25
Geormary completed assessment

Geormary:


main.js
```diff
@@ -0,0 +1,38 @@
+let values = { "⚀": 1, "⚁": 2, "⚂": 3, "⚃": 4, "⚄": 5, "⚅": 6 };
+let values1 = { 1: "⚀", 2: "⚁", 3: "⚂", 4: "⚃", 5: "⚄", 6: "⚅" };
+let die = ["⚀", "⚁", "⚂", "⚃", "⚄", "⚅"];
```
jabyess, line 3: I like your idea here of building different objects and array for different things! It looks like you're only actually using `die` though, which is great. I'm guessing you started with different ideas then figured out you only needed one of them.

main.js
```diff
@@ -0,0 +1,38 @@
...
+  let dicePara = document.getElementById("dice-para");
+  let button = document.getElementById("dice-roll");
+  let sumPara = document.getElementById("sum-para");
+  let historyList = document.getElementById("history-list")
+  button.addEventListener("click", (e) => {
+    let diceStr = "";
+    let sum = 0;
+    for (let i = 0; i < inputNum.value; i++) {
+      let randNum = Math.floor(Math.random() * (die.length));
+      diceStr += die[randNum];
```
jabyess, line 16: Love this! So simple, just using the random index to get the value out of the array.

main.js
```diff
@@ -0,0 +1,38 @@
...
+      console.log(randNum)
+      sum += (randNum+1)
+    }
+    dicePara.textContent = `Dice: ${diceStr}`;
+    sumPara.textContent = `Sum = ${sum}`;
```
jabyess, line 21: This looks like it should work! But where is your sum variable defined?

jabyess, line 21: This looks like it should work! But the test is failing and it's not obvious to me why that is just by looking at it right now.

main.js
Geormary: Yeah, I  thought an object would be the best way to go but struggled to figure out how to go about it. In the end, I  stuck with arrays and didnt have time to delete the other two

PR #28, opened 2021-02-25
Midmodule Assessment 

ftwash:


index.html
```diff
@@ -0,0 +1,69 @@
...
+  <head>
+    <link rel="stylesheet" href="./index.css">
+    <title>🎲Welcome to the Dice Roller🎲 </title>
+  </head>
+  <body>
+    <h1 id = "top_heading">🎲Welcome to the Dice Roller🎲</h1>
+    <h2 id = "welcome_para">Welcome! Please enter the number of dice you'd like to use:</h2>
+
+    <input id="number-input" type="number" placeholder="Number of dice"><br><br>
+    <button onclick = "rollDice()" id="mirror-button" type="submit">Roll 'em</button>
```
jabyess, line 12: Wondering why you chose to use the onclick attribute here instead of an eventListener on the document? Either way works though!

index.html
```diff
@@ -0,0 +1,69 @@
...
+    
+    <script>
+
+        const rollDice = () =>{
+    
+          let numOfRolls = document.getElementById("number-input").value
+          let dice = ['\u2680', '\u2681', '\u2682', '\u2683', '\u2684', '\u2685']
```
jabyess, line 19: How did you figure out to put the \u in front of each character here? I didn't even know about this! Very cool

index.html
```diff
@@ -0,0 +1,69 @@
...
+          for(i=0; i<numOfRolls; i++){
+            let randomRoll = dice[Math.floor(Math.random() * dice.length)]
+            //create a string containing all of the dice results 
+            let rollResult = 
+            let rollSum =
```
jabyess, line 24: You're on the right track here. How do you add up all the rolls to the sum?

index.html
ftwash: That was just sort of fresh in my mind from a lab I had been working on, and I was/am less acquainted with eventListener. I need more practice with that.

index.html
ftwash: google/ stack overflow! but I guess it was unnecessary because values were already provided for the die. Unnecessary but cool.

index.html
ftwash: figuring out the answer to this is one of my goals for this morning.

PR #29, opened 2021-02-25
Mid Module

AlekiChrome:


mikeboyle:
closing and reopening to run test automation, apologies for this

index.html
```diff
@@ -0,0 +1,75 @@
...
+    <link rel="stylesheet" href="styles.css">
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Document</title>
+</head>
+<body>
+    <h1>Dice Roller</h1>
+
+    Number of Dice: <input type="text" id="dice-number" />
```
myrasmith, line 14: Are there any HTML tags you could use to encapsulate the 'Number of Dice:' text?

index.html
```diff
@@ -0,0 +1,75 @@
...
+    <input type="button" onclick="doRoll('dice-number', 'dice-roll-result')" value="Roll the Dice!" />
```
myrasmith, line 15: Consider using a form to group related inputs, labels, and submit buttons!

index.html
```diff
@@ -0,0 +1,75 @@
...
+
+    <div>
+        sum =
+        <span style="font-size: 64px; line-height: normal" id="dice-roll-result"><span>
+    </div>
```
myrasmith, line 20: The tests rely on id='dice-para' and id='sum-para' being added to the right tags - looks like they are commented out below!

index.js
```diff
@@ -0,0 +1,60 @@
+const roll = (x) => {
+    let dice = ['\u2680', '\u2681', '\u2682', '\u2683', '\u2684', '\u2685'];
+    let results = '';
+    for (let i = 0; i < x; i++) {
+        results += dice[Math.floor(Math.random()*6)] + ' ';
+    }
+    return results;
+}
+
+const doRoll = (amount, result) => {
```
myrasmith, line 10: Since the id of the input tag and the results tag won't every change, it might not be necessary to have these as function parameters.

index.js
```diff
@@ -0,0 +1,60 @@
+const roll = (x) => {
```
myrasmith, line 1: Why did you decide to separate roll and doRoll into two functions?

index.js
```diff
@@ -0,0 +1,60 @@
...
+    let dice = ['\u2680', '\u2681', '\u2682', '\u2683', '\u2684', '\u2685'];
+    let results = '';
+    for (let i = 0; i < x; i++) {
+        results += dice[Math.floor(Math.random()*6)] + ' ';
+    }
+    return results;
```
myrasmith, line 7: Nice approach to creating a string of random dice that is the proper length.

myrasmith (commented):
Nice job creating the random string of dice! Some ways I think you could improve your code:
- Include HTML that is structured correctly for cypress to run its tests
- Introduce function parameters if they represent values that might be different depending on when you call the function, otherwise they are probably unnecessary
- consider combining two functions into one if one of the functions doesn't add much functionality.

PR #30, opened 2021-02-25
mid mod commit

jamee-maker:


index.html
```diff
@@ -0,0 +1,41 @@
...
+        border: solid 4px;
+      }
+    </style>
+  </head>
+  <body>
+    <h1>Dice Roller</h1>
+    <label for="">Number of dice:</label>
+    <input type="number" id="diceinput" />
+    <br />
+    <button onclick="data()" id="roll" type="submit">Roll the dice!</button>
```
jabyess, line 24: Wondering why you chose to use the `onclick` attribute here instead of using an event listener with `DOMContentLoaded`. Either way works though!

main.js
```diff
@@ -0,0 +1,43 @@
...
+const data = () => {
+  const input = Number(document.getElementById("diceinput").value);
+  let diceObj = {
+    "&#9856": 1,
+    "&#9857": 2,
+    "&#9858": 3,
+    "&#9859": 4,
+    "&#9860": 5,
+    "&#9861": 6,
+  };
```
jabyess, line 13: Clever usage of the object here! Love how you're using this to add the sum for each die.

main.js
```diff
@@ -0,0 +1,43 @@
...
+  let ul = document.getElementById("history");
+  let li = document.createElement("li");
+  let newArr = [];
+  for (let i = 1; i <= input; i++) {
+    let roll = diceRoll();
+    dicePara.innerHTML += roll;
+    newArr.push(roll);
+  }
+  for (let i = 0; i <= input; i++) {
+    if (diceObj[newArr[i]]) {
```
jabyess, line 29: If you started your `for()` loop at 1, you could skip using the if statement here. This totally works though.

jamee-maker: Interesting! What do you mean by that Jimmy? I have to use the first index to compare the values. Do you mean offsetting the position of my dices in my newArr?

index.html
jamee-maker: Ugh!!!!!! I know this was so frustrating!!!!! Great spot Jimmy! For some odd reason my event listeners where acting weird on that day, so for the sake of passing the test and finishing on time I chose this route. I understand how to use event listeners for the most part. It was just acting weird on the html page for some reason and I have no idea why but I can refactor my code now that we have time to see why.

main.js
jamee-maker: Thanks Jimmy!!! The moment I read the question I knew I would need an object perhaps to link a string with a number value. I also figured I could offset the dice html in the array so that each dice corresponds with the index in that array. and add the indexes.

main.js
```diff
@@ -0,0 +1,43 @@
...
+    "&#9857": 2,
+    "&#9858": 3,
+    "&#9859": 4,
+    "&#9860": 5,
+    "&#9861": 6,
+  };
+  if (input === 0) {
+    let p = document.getElementById("error");
+    p.textContent = "Error! Please enter a number.";
+  }
```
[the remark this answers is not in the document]
jamee-maker: Thanks! Just an embellishment haha

PR #32, opened 2021-02-25
work

alexanderl96-hub:


index.js
```diff
@@ -0,0 +1,62 @@
...
+        if(input.value !== ""){
+            const dice = ['⚀',
+                '⚁',
+                '⚂',
+                '⚃',
+                '⚄',
+                '⚅',]
+            arr = []
+            for(i = 0; i < input.value; i++){
+                let newArr = dice[Math.floor(Math.random()* dice.length)]
```
jabyess, line 20: This is a super clever way of getting a random dice value! nice work!

index.js
```diff
@@ -0,0 +1,62 @@
...
+                if(arr.length = 1){
+                    p2.textContent = arr
+                }else{
+                    p2.textContent = arr[newArr]
+                }
+                
+
+                if(p2.textContent === '⚀'){
+                    p.textContent = `Sum = ${1}`
+                    p.textContent.length.value
```
jabyess, line 31: What is this line doing?

index.js
```diff
@@ -0,0 +1,62 @@
...
+                }else if(p2.textContent === '⚂'){
+                    p.textContent = `Sum = ${3}`
+                }else if(p2.textContent === '⚃'){
+                    p.textContent = `Sum = ${4}`
+                }else if(p2.textContent === '⚄'){
+                    p.textContent = `Sum = ${5}`
+                }else if(p2.textContent === '⚅'){
+                    p.textContent = `Sum = ${6}`
+                
+                }
```
jabyess, line 43: Is there another way to re-write all of these if/else statements to get a sum? Maybe like using the index of the array?

index.js
```diff
@@ -0,0 +1,62 @@
...
+            const dice = ['⚀',
+                '⚁',
+                '⚂',
+                '⚃',
+                '⚄',
+                '⚅',]
+            arr = []
+            for(i = 0; i < input.value; i++){
+                let newArr = dice[Math.floor(Math.random()* dice.length)]
+                arr += newArr
```
jabyess, line 21: If `arr` is an array and `newArr` is an array, what is `+=` supposed to be doing here? Do you mean to use .push instead?

index.js
```diff
@@ -0,0 +1,62 @@
...
+                console.log(arr)
+                
+            }
+            const li = document.createElement('li')
+                li.textContent = arr
+                ul.appendChild(li)
+                if(p.textContent === li.textContent){
+                    return 
+                }else{
+                    li.textContent = `${arr} ${p.textContent}`
```
jabyess, line 53: I think this code won't work because `arr` is an array. and you can't put an array inside of a string like this - how would you convert the array to a string?

index.js
alexanderl96-hub: Thanks, I really appreciate this. It will be a much better way for me to get to improve my work.

index.js
alexanderl96-hub: sorry , but i'm a little bit confuse. you mean the last one? if is that one ,I forget to remove it from the code when I submit my work. I was working in different approach to solved the lab.

index.js
alexanderl96-hub: yes definitely. I was trying with .forEach(()) and for(let _ of_), but i got stuck. maybe i forget to put something.

index.js
alexanderl96-hub: well, i thought that it was doing the right thing but clear not

index.js
alexanderl96-hub: .split()

PR #34, opened 2021-02-26
raymond udeogu

rayu117:


index.js
```diff
@@ -0,0 +1,87 @@
...
+        // use map OR forEach callback and use if statements to reassign each number to a possible dice image
+        //another method is to make an object that assigns each number in the array to a corresponding object position => roll1: &#9856 or roll1: "&#9856"
+//i want to add dice values and make the value appear onscreen
+//i want to create a variable that stores the entire roll which equals the dice and their sum
+//each time i roll i want to store the dice rolls in paragraph
+//append each roll to history
+    //check number of clicks 
+        //only append roll when number of clicks is greater than 1
+        //get the dice roll and the sume and make a variable to represent them
+            //element must be li
```
jabyess, line 27: I love that you wrote out all your pseudocode here! 

One thing I remember you asked about in our 1:1 was whether you should write out everything before you started coding, since it took you a long time to get there. Sometimes I think this is a good strategy, but if you find yourself feeling stuck, it can also help to just start coding. While coding you may figure out what you need to do next.

Try it different ways, figure out what works for you and your process!

index.js
```diff
@@ -0,0 +1,87 @@
...
+            return myDice.rollFive
+        } else if (die === 6) {
+            return myDice.rollSix
+        }
+    }
+        viewList.push(makeDice(newDice))
+
+    
+    diceList.push(newDice)
+    userRoll--
```
jabyess, line 57: Cool that you used a `while()` loop and then manually decremented the counter! Don't forget you can also use a `for()` loop and count backwards

index.js
```diff
@@ -0,0 +1,87 @@
...
+    console.log(diceList)
+    //i want to take the number value of each element in the object/array and change it to the right html entity value
+    let view = document.getElementById('dice-para')
+
+    view.innerHTML = viewList
```
jabyess, line 62: At this point, viewList is an array. I don't think you can set .innerHTML to an array, it needs to be a string. How could you convert this array into a string?

index.js
```diff
@@ -0,0 +1,87 @@
...
+
+    let diceSum = diceList.reduce((acc,el) => {
+        return acc + el
+    })
```
jabyess, line 66: Excellent usage of reduce!!

index.js
```diff
@@ -0,0 +1,87 @@
...
+            return myDice.rollTwo
+        } else if (die === 3) {
+            return myDice.rollThree 
+        } else if (die === 4) {
+            return myDice.rollFour
+        } else if (die === 5) {
+            return myDice.rollFive
+        } else if (die === 6) {
+            return myDice.rollSix
+        }
```
jabyess, line 51: You're using an object to store all the dice string values, which is totally cool! What if you used an array instead? How could this if/else part be re-written?

PR #35, opened 2021-02-28
Shoaib Dar

SDAR30:


index.html
```diff
@@ -0,0 +1,39 @@
...
+
+<body>
+    <h1>Dice Roller</h1>
+
+    <form id="dice-form">
+        Number of Dice:
+        <input type="number" min="0" id="dice-input" required>
+        <br>
+        <button type="submit" id="dice-button">Roll the dice!</button>
+    </form>
```
Sparkadelic177, line 21: we could of also placed this inside of a section tag

index.js
```diff
@@ -0,0 +1,42 @@
...
+        addToHistory();
+    }
+
+    totalDiceRolls++;
+    const diceInput = document.querySelector("#dice-input").valueAsNumber;
+    let nums = []
+    let total = 0;
+    let str = "";
+    for (let i = 0; i < diceInput; i++) {
+        let diceRoll = Math.ceil(Math.random() * 6)
```
Sparkadelic177, line 32: instead of using the hard coded 6 we can us the length of the array `diceImages`

index.js
```diff
@@ -0,0 +1,42 @@
...
+        nums.push(diceRoll)
+        str += diceImages[diceRoll - 1]
+        total += diceRoll
+    }
+    dicePara.style.fontSize = "xx-large"
```
Sparkadelic177, line 37: we do this in the css file

index.js
```diff
@@ -0,0 +1,42 @@
+
+const diceImages = ["⚀", "⚁", "⚂", "⚃", "⚄", "⚅"]
+let totalDiceRolls = 0;
+
+const rollDice = document.querySelector("#dice-form");
+const dicePara = document.querySelector("#dice-para");
+const sumPara = document.querySelector("#sum-para");
+
+
+const addToHistory = () => {
```
Sparkadelic177, line 10: I like that you moved this function to its own thing

PR #36, opened 2021-02-28
Jailene DeVine-Jones

JDJ97:


index.html
```diff
@@ -0,0 +1,27 @@
...
+    <link rel="stylesheet" href="index.css">
+    <title>Document</title>
+</head>
+<body>
+    <form>
+    <h1>Dice Roller</h1>
+    <p>Number of dice: <input type="number" id="num-dice"> </p>
+    <button type="submit" id="roll-btn">Roll the dice!</button>
+    </form>
+    <div>
```
coreyladovsky, line 16: use semantic html prefer section to div

index.html
```diff
@@ -0,0 +1,27 @@
...
+    <p id="dice-para"></p>
+    <p id="sum-para">Sum = </p>
+    </div>
+    <div>
+    <h2>History</h2>
+    <ul>
+
+    </ul>
+    </div>
+</body>
```
coreyladovsky, line 26: Use a code formatter like prettier so make code more readable. option + shift + f

index.html
```diff
@@ -0,0 +1,27 @@
...
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <script src="index.js" defer></script>
+    <link rel="stylesheet" href="index.css">
+    <title>Document</title>
+</head>
+<body>
+    <form>
+    <h1>Dice Roller</h1>
+    <p>Number of dice: <input type="number" id="num-dice"> </p>
```
coreyladovsky, line 13: Use a label instead of a p here and either wrap it around the input or point it at the input id with the label `for` attribute

index.js
```diff
@@ -0,0 +1,43 @@
+const num = document.querySelector('#num-dice')
+const button = document.querySelector('#roll-btn')
+const diceDisplay = document.querySelector('#dice-para')
+const sumDisplay = document.querySelector('#sum-para')
+const ul = document.querySelector('ul')
+
+
+diceArr = ['&#9856', '&#9857', '&#9858', '&#9859', '&#9860', '&#9861']
```
coreyladovsky, line 8: Remember to declare all variables

index.js
```diff
@@ -0,0 +1,43 @@
...
+
```
coreyladovsky, line 9: should listen for "submit" on form instead of button "click".

index.js
```diff
@@ -0,0 +1,43 @@
...
+button.addEventListener('click', (event) => {
+    event.preventDefault()
+    sumDisplay.textContent="Sum = "
+    let newDiceArr = []
```
coreyladovsky, line 13: could be const b/c not reassigned

index.js
```diff
@@ -0,0 +1,43 @@
...
+const diceDisplay = document.querySelector('#dice-para')
+const sumDisplay = document.querySelector('#sum-para')
+const ul = document.querySelector('ul')
+
+
+diceArr = ['&#9856', '&#9857', '&#9858', '&#9859', '&#9860', '&#9861']
+
+button.addEventListener('click', (event) => {
+    event.preventDefault()
+    sumDisplay.textContent="Sum = "
```
coreyladovsky, line 12: space around = should be present

index.js
```diff
@@ -0,0 +1,43 @@
...
+    let newDiceArr = []
+    for (let i = 0; i < num.value; i++){
```
coreyladovsky, line 14: .valueAsNumber for slightly cleaner code

index.js
```diff
@@ -0,0 +1,43 @@
...
+        newDiceArr[i] = diceArr[Math.floor(Math.random() * diceArr.length)]
```
coreyladovsky, line 15: Better to push into array

index.js
```diff
@@ -0,0 +1,43 @@
...
+        }else if (newDiceArr[i] === '&#9859'){
+            sum += 4
+        }else if (newDiceArr[i] === '&#9860'){
+            sum += 5
+        } else {
+            sum += 6
+        }
+
+    }
+    let diceHistoryList = document.createElement('li')
```
coreyladovsky, line 35: const

index.js
```diff
@@ -0,0 +1,43 @@
...
+    event.preventDefault()
+    sumDisplay.textContent="Sum = "
+    let newDiceArr = []
+    for (let i = 0; i < num.value; i++){
+        newDiceArr[i] = diceArr[Math.floor(Math.random() * diceArr.length)]
+    }
+    
+    let sum = 0
+    for(let i = 0; i < newDiceArr.length; i++){
+        if (newDiceArr[i] === '&#9856'){
```
coreyladovsky, line 20: code can be simplified by adding sum during first loop. We can get the die value by getting the index + 1

PR #37, opened 2021-02-28
all tests passed 

itsanasna:


index.html
```diff
@@ -0,0 +1,31 @@
...
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <script src="./index.js" defer></script>
+    <title>Dice Rolling</title>
+    
+
+    <section id="section-one">
```
coreyladovsky, line 11: Nice semantic tags

index.html
```diff
@@ -0,0 +1,31 @@
...
+        <form >
+            Dice rolls:<input type="number" placeholder="Number of rolls" id="number-of-rolls">
```
coreyladovsky, line 13: Use a label here and either wrap it around the input or point it at the input id with the label `for` attribute

index.js
```diff
@@ -0,0 +1,67 @@
...
+    '&#9856',
+    '&#9857',
+    '&#9858',
+    '&#9859',
+    '&#9860',
+    '&#9861'
+]
+
+
+button.addEventListener('click', (event) => {
```
coreyladovsky, line 16: should listen for "submit" on form instead of button "click".

coreyladovsky, line 16: Consider pulling out these callback functions into named functions so that code can be a bit more modular.

index.js
```diff
@@ -0,0 +1,67 @@
...
+    event.preventDefault()
+    total.textContent="Sum = "
```
coreyladovsky, line 18: space around = should be present

index.js
```diff
@@ -0,0 +1,67 @@
...
+    let array = []
```
coreyladovsky, line 19: could be const b/c not reassigned

index.js
```diff
@@ -0,0 +1,67 @@
...
+    for (let i = 0; i < numInput.value; i++){
```
coreyladovsky, line 20: .valueAsNumber for slightly cleaner code

index.js
```diff
@@ -0,0 +1,67 @@
...
+        array[i] = dices[Math.floor(Math.random() * 6)]
```
coreyladovsky, line 21: Better to push into array

index.js
```diff
@@ -0,0 +1,67 @@
...
+    }
+    let sum = 0
+    for(let i = 0; i < array.length; i++){
```
coreyladovsky, line 24: code can be simplified by adding sum during first loop. We can get the die value by getting the index + 1

index.js
```diff
@@ -0,0 +1,67 @@
...
+            sum += 3 
+        }else if (array[i] === '&#9859'){
+            sum += 4
+        }else if (array[i] === '&#9860'){
+            sum += 5
+        } else if ( array[i] === '&#9861') {
+            sum += 6
+        }
+    }
+    let list = document.createElement('li')
```
coreyladovsky, line 39: const

PR #38, opened 2021-02-28
Complete Assessment - Jeffrey Sebastian

JeffreySebastian-Pursuit:


index.js
```diff
@@ -0,0 +1,38 @@
+const diceArr = [
```
coreyladovsky, line 1: I like this but how would structuring your dice as an object like: 
{1: "⚀", 2: "⚁" .... etc } improve your code?

index.html
```diff
@@ -0,0 +1,25 @@
...
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Document</title>
+    <script src="index.js" defer></script>
+    <link rel="stylesheet" href="styles.css" />
+  </head>
+  <body>
+    <h1>Dice Roller</h1>
+    <div>
```
coreyladovsky, line 13: use semantic html prefer section to div

index.js
```diff
@@ -0,0 +1,38 @@
...
+];
+
+const inputDice = document.querySelector("#inputDice");
+const diceButton = document.querySelector("#diceButton");
+const outputDice = document.querySelector("#dice-para");
+const sumOfDice = document.querySelector("#sum-para");
+let rollDice = "";
+let sum = 0;
+
+diceButton.addEventListener("click", () => {
```
coreyladovsky, line 17: Don't make an anonymous function just to call a named function. Instead put the named function only. Like:
diceButton.addEventListener("click", randomDice)

styles.css
```diff
@@ -0,0 +1,16 @@
+body{
+    background-color: black;
+    color: white;
+    display: flex;
```
coreyladovsky, line 4: Love seeing you using flex. Nice!

index.html
```diff
@@ -0,0 +1,25 @@
...
+      Number Of Dice<input id="inputDice" type="number" />
+      <button type="button" id="diceButton">Roll the Dice</button>
```
coreyladovsky, line 15: Best practice is to use forms around inputs and buttons

index.js
```diff
@@ -0,0 +1,38 @@
...
+  randomDice();
+});
+
+const randomDice = () => {
+  rollDice.innerHTML = "";
```
coreyladovsky, line 22: rollDice is a string so does not have an innerHTML property.

index.js
```diff
@@ -0,0 +1,38 @@
...
+  for (let i = 0; i < inputDice.value; i++) {
+    let newDice = [Math.floor(Math.random() * diceArr.length)];
```
coreyladovsky, line 24: could be const

index.js
```diff
@@ -0,0 +1,38 @@
...
+    rollDice += diceArr[newDice].dice;
+    sum += diceArr[newDice].value;
+  }
+  sumOfDice.textContent = `Sum = ${sum}`;
+  outputDice.textContent = rollDice;
+  inputDice.value = "";
+};
+
+diceButton.addEventListener("click", () => {
```
coreyladovsky, line 33: create separate function and call it from randomDice instead of having two click listeners.

index.js
```diff
@@ -0,0 +1,38 @@
...
+const history = document.querySelector("#diceHistory");
+let li = document.createElement("li");
```
coreyladovsky, line 35: const

index.js
```diff
@@ -0,0 +1,38 @@
...
+li.textContent = `${rollDice} = ${sum} `;
```
coreyladovsky, line 36: I think this might be happening after each element is made instead of being one behind. Think about adding some more logic or outside counter to know if should be appended yet or not.

PR #40, opened 2021-02-28
Jordan Bobadilla

jordanbobadilla:


index.html
```diff
@@ -0,0 +1,31 @@
+<!DOCTYPE html>
+<html lang="en">
```
coreyladovsky, line 2: Nice semantic tags

index.html
```diff
@@ -0,0 +1,31 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Roll The Dice</title>
+    <link rel="stylesheet" href="./style.css">
+    <script src="./index.js" defer></script>
+</head>
+<body>
+    <header>
+        <h1> Dice Roller</h1>
+        <form>
+            <label>Number of dice:</label>
```
coreyladovsky, line 15: labels should either wrap inputs or point to the input it's relating to. Needs the for attribute and points to input id.

index.js
```diff
@@ -0,0 +1,30 @@
+const diceRoll = ["⚀","⚁","⚂","⚃","⚄","⚅"]
```
coreyladovsky, line 1: I like this

index.js
```diff
@@ -0,0 +1,30 @@
...
+
+const input = document.querySelector("input")
+const button = document.querySelector("button")
```
coreyladovsky, line 4: We don't use this variable for anything, so this line should be removed

index.js
```diff
@@ -0,0 +1,30 @@
...
+const dice = document.querySelector("#dice-para")
+const sum = document.querySelector("#sum-para")
+const ul = document.querySelector("ul")
+const form = document.querySelector("form")
+
+form.addEventListener("submit", (e) => {
```
coreyladovsky, line 10: Consider pulling out these callback functions into named functions so that code can be a bit more modular.

index.js
```diff
@@ -0,0 +1,30 @@
...
+    e.preventDefault()
+    let roll = []
```
coreyladovsky, line 12: could be const b/c not reassigned

index.js
```diff
@@ -0,0 +1,30 @@
...
+    let sumVal = 0
+    sum.textContent = "Sum ="
+    for (let i = 0; i < Number(input.value); i++) {
```
coreyladovsky, line 15: .valueAsNumber for slightly cleaner code

PR #41, opened 2021-02-28
Coreen Cooper

CoreenCooper:


index.js
```diff
@@ -0,0 +1,43 @@
+const dice = [
+    { die:"&#9856", value: 1},
+    { die:"&#9857", value: 2},
+    { die:"&#9858", value: 3},
+    { die:"&#9859", value: 4},
+    { die:"&#9860", value: 5},
+    { die:"&#9861", value: 6},
+]
```
Sparkadelic177, line 8: I like the use of both array and object

index.js
```diff
@@ -0,0 +1,43 @@
...
+history()
+}
+
+const ul = document.querySelector("ul");
+
+const history = () => {
+    const listItem = document.createElement("li");
+    listItem.textContent = displayDice.textContent;
+    ul.appendChild(listItem);
+}
```
Sparkadelic177, line 40: I like that you destructed this logic in its own function

index.js
```diff
@@ -0,0 +1,43 @@
...
+let sum = 0;
+let rollCount = 0;
+
+const diceRoller = (event) => {
+    event.preventDefault();
+    rollResult = "";
+    sum = 0;
+    const userInput = Number(numOfDice.value);
+    for(let i = 0; i < userInput; i++) {
+        const randNum = Math.floor(Math.random() * 6);
```
Sparkadelic177, line 24: instead of using the hard coded value, we can also use the length of `dice`

index.js
```diff
@@ -0,0 +1,43 @@
...
+const displaySum = document.querySelector("#sum-para"); // display sum result
+let rollResult = "";
+let sum = 0;
+let rollCount = 0;
+
+const diceRoller = (event) => {
+    event.preventDefault();
+    rollResult = "";
+    sum = 0;
+    const userInput = Number(numOfDice.value);
```
Sparkadelic177, line 22: we can use `.valueAsNumber` instead of .value , it will be slightly cleaner

index.js
```diff
@@ -0,0 +1,43 @@
...
+    for(let i = 0; i < userInput; i++) {
+        const randNum = Math.floor(Math.random() * 6);
+        rollResult += dice[randNum]["die"];
+        sum += dice[randNum]["value"];
```
Sparkadelic177, line 26: super clean !

index.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+rollCount++
+displayDice.innerHTML = rollResult;
+displaySum.textContent = `Sum = ${sum}`;
+history()
```
Sparkadelic177, line 31: To append the history subsequently, we can create a some logic  to check to see if there is already something in the `display.innerHTML` before giving it its value. If so then add the current `display.innerHTML`  in the history then add the new `display.innerText` value.

index.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+const ul = document.querySelector("ul");
```
Sparkadelic177, line 34: we can place this outside of the event listener, we don't need to get a reference to this element for event submit

PR #42, opened 2021-02-28
Yashiry A Complete Assessment 

yashiazize:


diceRolling.js
```diff
@@ -0,0 +1,31 @@
...
+  { image: "&#9861", number: 6 },
+];
+
+const form = document.querySelector("#dice-roll-form");
+
+
+form.addEventListener("submit", (event) => {
+  event.preventDefault();
+  const display = document.querySelector("#dice-para");
+  const input = document.querySelector("#dice-amount").valueAsNumber;
```
Sparkadelic177, line 16: nice, I liked that you used the valueAsNumber property !!!

diceRolling.js
```diff
@@ -0,0 +1,31 @@
...
+  let randomDiceDisplay = "";
+  let sumTotal = 0;
+  for (let i = 0; i < input; i++) {
+    let randomDice = dice[Math.floor(Math.random() * dice.length)]; // create random dice
+    randomDiceDisplay += randomDice.image;
+    sumTotal += randomDice.number;
+  }
+  display.innerHTML = randomDiceDisplay;
+  sumDisplay.textContent = `Sum = ${sumTotal}`;
+    const ul = document.querySelector("ul");
```
Sparkadelic177, line 27: We don't have to select the ul for every form submission, but this still works

diceRolling.js
```diff
@@ -0,0 +1,31 @@
...
+    const li = document.createElement("li");
+    ul.appendChild(li);
+    li.textContent = `${display.innerHTML} = ${sumTotal}`;
```
Sparkadelic177, line 30: To append the history subsequently, we can create a separate function to check to see when if there already something in the `display.innerText` If so then add whats current there in the history then add the new `display.innerText` value

PR #43, opened 2021-02-28
Poonam 

PoonamDass:


index.html
```diff
@@ -0,0 +1,32 @@
...
+</head>
+<body>
+    <h1>Dice Roller</h1>
+
+    <section >
+        <label>Number of Dice:</label> 
+        <input id="number-of-dice" type="number" min=1 required>
+        <br>
+        <button id="roll-dice-btn" type="submit">Roll the dice!</button>
+    </section>
```
Sparkadelic177, line 19: Don't forget to add the ,`form` tag

index.js
```diff
@@ -0,0 +1,29 @@
+//const dices = ["&#9856", "&#9857", "&#9858", "&#9859", "&#9860", "&#9861"]
+const rollDiceBtn = document.querySelector("#roll-dice-btn")
+const numberOfDice = document.querySelector("#number-of-dice")
+const showRoll = document.querySelector("#dice-para")
+const sumOfRolls= document.querySelector("#sum-para")
+
+const dices =[{DiceNum:1,Dice: "&#9856"}, {DiceNum:2, Dice:"&#9857"}, {DiceNum:3, Dice:"&#9858"}, {DiceNum:4,Dice: "&#9859"}, {DiceNum:5, Dice:"&#9860"}, {DiceNum:6, Dice:"&#9861"}]
```
Sparkadelic177, line 7: Nice use of an array and object

index.js
```diff
@@ -0,0 +1,29 @@
...
+    let sum=0
+    for(let i=0; i<numberOfDice.value; i++){
+        let random= dices[Math.floor(Math.random() * dices.length)]
+        showRoll.innerHTML += random["Dice"]
+
+        sum+= Number(random["DiceNum"])
+        sumOfRolls.textContent = `Sum: ${sum}`
+    }
+
+    const ul= document.querySelector("#list")
```
Sparkadelic177, line 21: We don't need to grab the `ul` here for every form submission. This can be done just once

index.js
```diff
@@ -0,0 +1,29 @@
...
+    const li= document.createElement("li")
+    li.textContent = `${showRoll.innerHTML}=${sum}`
+    ul.appendChild(li)
```
Sparkadelic177, line 24: To append the history subsequently, we can create a some logic to check to see if there is already something in the `showRoll.innerHTML` before giving it its value. If so then add the current `showRoll.innerHTML`  in the history then add the new `showRoll.innerHTML` value.

PR #45, opened 2021-02-28
completed assessment

AdamTahiri:


dice.js
```diff
@@ -0,0 +1,29 @@
+// const dice = ["&#9856", "&#9857", "&#9858", "&#9859", "&#9860", "&#9861"];
+const dice = ["⚀", "⚁", "⚂", "⚃", "⚄", "⚅"];
```
coreyladovsky, line 2: Nice, I like this a lot

index.html
```diff
@@ -0,0 +1,28 @@
...
+  </head>
+  <body>
+    <h1>Dice Roller</h1>
+
+    Number of dice:
+    <form>
+      <input type="number" id="die-number-input" />
+      <button id="die-button" type="submit">Roll the dice!</button>
+    </form>
+    <section>
```
coreyladovsky, line 17: Nice semantic tags

dice.js
```diff
@@ -0,0 +1,29 @@
...
+const input = document.querySelector("#die-number-input");
+const showDice = document.querySelector("#dice-para");
+const sumOfDice = document.querySelector("#sum-para");
+const historyUl = document.querySelector("ul");
+const form = document.querySelector("form");
+
+form.addEventListener("submit", (event) => {
+  event.preventDefault();
+  sumOfDice.textContent = "Sum = "
+  sum = 0;
```
coreyladovsky, line 13: remember to declare variables. 
let sum

dice.js
```diff
@@ -0,0 +1,29 @@
...
+  displayDice = "";
```
coreyladovsky, line 14: LET

dice.js
```diff
@@ -0,0 +1,29 @@
...
+  for (let i = 0; i < input.valueAsNumber; i++) {
+    let randIdx = Math.floor(Math.random() * dice.length);
```
coreyladovsky, line 16: could be const

dice.js
```diff
@@ -0,0 +1,29 @@
...
+    displayDice += dice[randIdx];
+    sum += randIdx + 1;
+  }
+  showDice.textContent = displayDice;
+  sumOfDice.textContent = sumOfDice.textContent + sum;
+
+  if (input.valueAsNumber > 0) {
```
coreyladovsky, line 23: I don't think we care about input value here (but if you do add a min to your input). Instead you want to make sure that we're one behind with the history. 

Could create a variable that keeps track of submit counts. And make li's depending on that. Grab the text of the showDice before updating it to the new value.

PR #46, opened 2021-02-28
Jessica Netto

jessicanetto:


index.html
```diff
@@ -0,0 +1,33 @@
...
+<body>
+    
+    <form>
+        <div id="num-of-dice">Number of Dice:</div> 
+        <input type="number" id="enter-dice-num" placeholder= 0> 
+</br>
+</br>
+        <button type="reset" id="roll-button">Roll the dice!</button>
+    </form>
+    <div>
```
coreyladovsky, line 21: use semantic html prefer section to div

index.html
```diff
@@ -0,0 +1,33 @@
...
+    <link rel="stylesheet" href="index.css">
+    <title> Dice Roller</title>
+</head>
+<h1>Dice Roller</h1>
+<body>
+    
+    <form>
+        <div id="num-of-dice">Number of Dice:</div> 
+        <input type="number" id="enter-dice-num" placeholder= 0> 
+</br>
```
coreyladovsky, line 17: Be wary of using line break tags. Best practice is control position of elements via css

index.html
```diff
@@ -0,0 +1,33 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <script src="./index.js" defer></script>
+    <link rel="stylesheet" href="index.css">
+    <title> Dice Roller</title>
+</head>
+<h1>Dice Roller</h1>
+<body>
+    
+    <form>
+        <div id="num-of-dice">Number of Dice:</div>
```
coreyladovsky, line 15: Use a label instead of a div here and either wrap it around the input or point it at the input id with the label `for` attribute

index.html
```diff
@@ -0,0 +1,33 @@
...
+</head>
+<h1>Dice Roller</h1>
+<body>
+    
+    <form>
+        <div id="num-of-dice">Number of Dice:</div> 
+        <input type="number" id="enter-dice-num" placeholder= 0> 
+</br>
+</br>
+        <button type="reset" id="roll-button">Roll the dice!</button>
```
coreyladovsky, line 19: Be wary of type="reset" as it tends to annoy users. Typically type button is better, and in forms type submit is best.

index.js
```diff
@@ -0,0 +1,67 @@
+//put dice in array to pull from
+let dice = [
```
coreyladovsky, line 2: I like this but how would structuring your dice as an object like: 
{1: "&#9856", 2: "&#9857" .... etc } improve your code? 
Or putting the codes in an array and thinking about their values as their index + 1

index.js
```diff
@@ -0,0 +1,67 @@
+//put dice in array to pull from
```
coreyladovsky, line 1: Prefer const to let b/c not reassigned

index.js
```diff
@@ -0,0 +1,67 @@
...
+    value: 5,
+  },
+  {
+    di: "&#9861",
+    value: 6,
+  },
+];
+//grab from html
+let rollTheDiceButton = document.getElementById("roll-button");
+let enteredDiceNum = document.getElementById("enter-dice-num");
```
coreyladovsky, line 30: prefer const

index.js
```diff
@@ -0,0 +1,67 @@
...
+    di: "&#9860",
+    value: 5,
+  },
+  {
+    di: "&#9861",
+    value: 6,
+  },
+];
+//grab from html
+let rollTheDiceButton = document.getElementById("roll-button");
```
coreyladovsky, line 29: Use querySelector when possible, more current and up to date. Also consistent with code you write lower down

index.js
```diff
@@ -0,0 +1,67 @@
...
+const sumStorage = document.querySelector("#sum-para");
+
+//list
+const ul = document.createElement("ul");
+document.body.appendChild(ul);
+
+//execute function onclick
+rollTheDiceButton.addEventListener("click", () => {
+  rollTheDice();
+});
```
coreyladovsky, line 43: Don't make an anonymous function just to call a named function. Instead put the named function only. Like:
rollTheDiceButton.addEventListener("click", rollTheDice)


Also: should listen for "submit" on form instead of button "click", button will be type submit. This will fire the event. You're rollTheDice will then need to take in an Event Object as it's argument which you'll then need to called event.preventDefault()

index.js
```diff
@@ -0,0 +1,67 @@
...
+
+//make function with a loop to randomize rolls and store cumulative value
+function rollTheDice() {
+  diceElement.innerHTML = "";
+  let sum = 0;
+  //no need to roll multiple dice, same result as rolling one multiple times
+  //loop runs as many times as number of dice to simulate this
+  for (let i = 0; i < enteredDiceNum.value; i++) {
```
coreyladovsky, line 51: .valueAsNumber for slightly cleaner code

index.js
```diff
@@ -0,0 +1,67 @@
...
+    let generateRoll = Math.floor(Math.random() * 6);
```
coreyladovsky, line 52: const

index.js
```diff
@@ -0,0 +1,67 @@
...
+//execute function onclick
+rollTheDiceButton.addEventListener("click", () => {
+  rollTheDice();
+});
+
+//make function with a loop to randomize rolls and store cumulative value
+function rollTheDice() {
+  diceElement.innerHTML = "";
+  let sum = 0;
+  //no need to roll multiple dice, same result as rolling one multiple times
```
coreyladovsky, line 49: I like all these comments. Makes it easier to follow the logic. Thanks!

index.js
```diff
@@ -0,0 +1,67 @@
...
+  //loop runs as many times as number of dice to simulate this
+  for (let i = 0; i < enteredDiceNum.value; i++) {
+    let generateRoll = Math.floor(Math.random() * 6);
+
+    diceElement.innerHTML += dice[generateRoll].di;
+    sum += dice[generateRoll].value;
+  }
+  sumStorage.textContent = `The sum of this roll is = ${sum}`;
+  let rollCount = 0;
```
coreyladovsky, line 58: Each time this function runs rolllCount will be reinitialized to 0. And then immediately incremented by 1.

index.js
```diff
@@ -0,0 +1,67 @@
...
+  rollCount++;
+
+  //history
+  if (rollCount > 0) {
```
coreyladovsky, line 62: rollCount will always be greater than 0. We want the old text that was there, so we should be able to do most of this logic first. And then update the sum and dice string.

PR #48, opened 2021-02-28
Miguel Garcia

garcia13:


index.js
```diff
@@ -0,0 +1,43 @@
...
+let counter = 0;
+
+const diceRoller = (event) => {
+    event.preventDefault();
+    rollResult.textContent = "";
+    let num = numberOfDice.valueAsNumber;
+    let sum = 0;
+    for (let i = 0; i < num; i++) {
+        const diceIndex = Math.floor(Math.random() * dice.length);
+        rollResult.textContent += String.fromCodePoint(dice[diceIndex]);
```
Sparkadelic177, line 26: This is an interesting way of getting the dice image, we could of also just used the dice lookalike that was the readme

index.js
```diff
@@ -0,0 +1,43 @@
...
+        sum += diceIndex;
+    }
+    dieSum.textContent = `Sum: ${sum + num}`;
+    counter++;
+    newRollResult = `${rollResult.textContent} = ${sum + num}`
+    if (counter > 1) {        
+        const li = document.createElement("li");
+        li.textContent = oldResult;
+        ul.appendChild(li);
```
Sparkadelic177, line 35: Nice work, another way to append the history subsequently, we can create a some logic to check to see if there is already something in the ` rollResult.textContent` before giving it its value. If so then add the current ` rollResult.textContent`  in the history then add the new ` rollResult.textContent` value.

index.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+    oldResult = newRollResult
+    newRollResult = ""
+}
+
+
+
+form.addEventListener("submit", (diceRoller));
```
Sparkadelic177, line 43: no need for the `()` around `diceRoller` here

PR #49, opened 2021-02-28
Joshua Mendez

joshuafmendez:


index.html
```diff
@@ -0,0 +1,43 @@
...
+    Use the follow HTML entity values for dice: -->
+
+    <h1>Dice Roller</h1>
+    <form id='dice-form'>
+        Number of dice: <input id='number-of-dice' type="number">
+        <br>
+        <button type='submit'>Roll the dice!</button>
+    </form>
+
+    <section id='dice-sum-section'>
```
coreyladovsky, line 34: Nice semantic tags

index.html
```diff
@@ -0,0 +1,43 @@
...
+    The only li elements should be your history dice rolls
+    You should have exactly one input element where the number of dice can be entered
+    You should have exactly one button element that the user can click on to roll the dice
+    The die result from rolling should be stored in an element with the id "dice-para"
+    The sum from rolling should be stored in an element with the id "sum-para"
+    Use the follow HTML entity values for dice: -->
+
+    <h1>Dice Roller</h1>
+    <form id='dice-form'>
+        Number of dice: <input id='number-of-dice' type="number">
```
coreyladovsky, line 29: Use a label here and either wrap it around the input or point it at the input id with the label `for` attribute

index.html
```diff
@@ -0,0 +1,43 @@
...
+        <br>
```
coreyladovsky, line 30: Be wary of using line break tags. Best practice is control position of elements via css

index.js
```diff
@@ -0,0 +1,41 @@
+document.addEventListener('DOMContentLoaded', () => {
+    let dices = ["⚀", "⚁", "⚂", "⚃", "⚄", "⚅" ]
```
coreyladovsky, line 2: could be const b/c not reassigned

coreyladovsky: Seems like let have become your default declaration, switch it to const

index.js
```diff
@@ -0,0 +1,41 @@
...
+    let diceRollCount = 0;
+    let form = document.querySelector('#dice-form');
+    let dp = document.querySelector('#dice-para');
+    let sumP = document.querySelector('#sum-para')
+    form.addEventListener('submit', (event) => {
```
coreyladovsky, line 7: Consider pulling out these callback functions into named functions so that code can be a bit more modular.

index.js
```diff
@@ -0,0 +1,41 @@
...
+        event.preventDefault();
+        let input = document.querySelector('#number-of-dice');
+        if(diceRollCount > 0 && input.value !== ''){
```
coreyladovsky, line 10: could add a min on input in html

index.js
```diff
@@ -0,0 +1,41 @@
...
+            historySection();
```
coreyladovsky, line 11: I really like this call to a helper function

index.js
```diff
@@ -0,0 +1,41 @@
...
+        }
+        if(input.value !== ''){
```
coreyladovsky, line 13: could make input required in html

index.js
```diff
@@ -0,0 +1,41 @@
...
+            diceRollCount++;
+        }
+        let diceNum = Number(input.value);
```
coreyladovsky, line 16: .valueAsNumber for slightly cleaner code

index.js
```diff
@@ -0,0 +1,41 @@
...
+        list.appendChild(li)
+    }
+    
+    const randomizeAndDisplay = diceNum => {
+        let diceArr = [];
+        let sum = 0;
+        let showDice = '';
+        for(let i = 0; i < diceNum;i++){
+            let randomDice = Math.floor(Math.random() * 6) + 1;
+            diceArr.push(randomDice);
```
coreyladovsky, line 34: I'm not seeing diceArr being used anywhere else. Remove these lines and that variable, not needed.

index.js
```diff
@@ -0,0 +1,41 @@
...
+            showDice += dices[randomDice - 1];
+            sum += randomDice;
```
coreyladovsky, line 36: clean

style.css
```diff
@@ -0,0 +1,23 @@
+section {
+    border: 5px solid black;
+    margin-top: 20px;
+}
+
+#dice-sum-section {
+    height: 150px;
+    position: relative;
```
coreyladovsky, line 8: I'd love to see some flexbox styling :-)

PR #51, opened 2021-02-28
Carina Taveras

tCarina:


index.html
```diff
@@ -0,0 +1,30 @@
...
+  </head>
+  <body>
+    
+      <h1>Dice Roller</h1>
+      <form id = "dice-form">
+        Number Of Dice: <input type="number" id="dice-length" min="1"/>
+        <button type="submit">Roll The Dice!</button>
+      </form>
+      
+    <div>
```
coreyladovsky, line 20: prefer section to div

index.js
```diff
@@ -0,0 +1,39 @@
+const diceObj = [
```
coreyladovsky, line 1: I like this but how would structuring your dice as an object like: 
{1: "⚀", 2: "⚁" .... etc } improve your code?

index.js
```diff
@@ -0,0 +1,39 @@
...
+  { dice: "⚀", val: 1 },
+  { dice: "⚁", val: 2 },
+  { dice: "⚂", val: 3 },
+  { dice: "⚃", val: 4 },
+  { dice: "⚄", val: 5 },
+  { dice: "⚅", val: 6 },
+];
+const form = document.querySelector("#dice-form");
+let diceOutput = document.querySelector("#dice-para");
```
coreyladovsky, line 10: could be const

index.js
```diff
@@ -0,0 +1,39 @@
...
+let sumOutput = document.querySelector("#sum-para");
```
coreyladovsky, line 11: could be const

index.js
```diff
@@ -0,0 +1,39 @@
...
+const ul = document.querySelector("#dice-history");
+let sumDis = 0
+
+const rollDice = (event) => {
+  let sum = 0;
+  event.preventDefault();
```
coreyladovsky, line 17: Traditionally event.preventDefault() should be the first line of the function.

index.js
```diff
@@ -0,0 +1,39 @@
...
+  diceOutput.textContent = "";
+  const num = document.querySelector("#dice-length").value;
```
coreyladovsky, line 19: .valueAsNumber for slightly cleaner code

index.js
```diff
@@ -0,0 +1,39 @@
...
+  const diceNum = parseInt(num);
+  for (let i = 0; i < diceNum; i++) {
+    const ranDice = Math.floor(Math.random() * diceObj.length);
+    diceOutput.textContent += diceObj[ranDice].dice;
+    sum += diceObj[ranDice].val;
+    sumOutput.textContent = "Sum = " + sum;
```
coreyladovsky, line 25: could be done at end outside of for loop

index.js
```diff
@@ -0,0 +1,39 @@
...
+    sumDis = sum
+  }
+};
+
+form.addEventListener("submit", rollDice);
+
+form.addEventListener("submit", () => {
```
coreyladovsky, line 32: create separate function and call it from rollDice instead of having two submit listeners.

index.js
```diff
@@ -0,0 +1,39 @@
...
+  let li = document.createElement("li")
```
coreyladovsky, line 33: could be const

index.js
```diff
@@ -0,0 +1,39 @@
...
+  li.textContent =  `${diceOutput.textContent} = ${sumDis}`
+  ul.appendChild(li)
```
coreyladovsky, line 35: I think this might be happening after each element is made instead of being one behind. Think about adding some more logic or outside counter to know if should be appended yet or not.

PR #52, opened 2021-02-28
Marsha Hall

mphall77:


dice.js
```diff
@@ -0,0 +1,86 @@
+console.log("I am awesome!");
+
+diceArr = ["&#9856", "&#9857", "&#9858", "&#9859", "&#9860", "&#9861"];
+diceVal = ["1", "2", "3", "4", "5", "6"];
```
Sparkadelic177, line 4: we could combine these two by creating an array that holds the values in an object. 
`[{dice:"&9856", value:1}]`

dice.js
```diff
@@ -0,0 +1,86 @@
...
+let historyLog;
+let HistoryLog2;
+let clicks = 0;
+
+// FUNCTIONS
+
+const getDice = (event) => {
+	let rolls = userInput.valueAsNumber;
+	for (let i = 0; i < rolls; i++) {
+        let num = Math.floor(Math.random() * 5) + 1;
```
Sparkadelic177, line 25: instead of hard coding the 5 and adding the plus one, we could use the length of `diceArr`

dice.js
```diff
@@ -0,0 +1,86 @@
...
+        let die = diceArr[num].replace("&#", "");
+		diceList += String.fromCharCode(die);
```
Sparkadelic177, line 27: This is a interesting way of getting the die image. we could also use `.innerHTML` to get the die without needing replace the `&#` from the string

dice.js
```diff
@@ -0,0 +1,86 @@
...
+		sum += Number(diceVal[num]);
```
Sparkadelic177, line 28: You can also just declare the values inside of this array to be numbers instead of strings

dice.js
```diff
@@ -0,0 +1,86 @@
...
+	const createHistory = () => {
+		const li = document.createElement("li");
+		li.textContent = historyLog2;
+		ul.appendChild(li);
+		reset();
+	};
+
+	clicks++;
+	historyLog = `${pDice.textContent} = ${sum}`;
+	clicks <= 1 ? reset() : createHistory();
```
Sparkadelic177, line 51: Love this ternary operation

dice.js
```diff
@@ -0,0 +1,86 @@
...
+		sum = 0;
+		historyLog = "";
+	};
+
+	const createHistory = () => {
+		const li = document.createElement("li");
+		li.textContent = historyLog2;
+		ul.appendChild(li);
+		reset();
+	};
```
Sparkadelic177, line 47: I like that you separated the logic, super clean

index.html
```diff
@@ -0,0 +1,38 @@
...
+			<h2>Dice Roller</h2>
+		</header>
+
+		<!-- INPUT -->
+		<section>
+			<label for="userInput">
+				Number of dice:
+				<input type="number" name="userInput" min="1" />
+			</label>
+			<button id="roll">Roll the dice!</button>
```
Sparkadelic177, line 23: Don't forget the form tag

PR #53, opened 2021-02-28
Done

bedonjonathan:
The test was not checking to see if the history list was generated after the second click.

### DIFF
--- a/dice.css
+++ b/dice.css
@@ -1,17 +1,20 @@
-header {
-    display: flex;
-    ;
-}
-section {
+
+section ,header ,form {
     border: navy 2px solid;
     padding: 1%;
     margin: 2.5%  ;
+    border-radius: 5%;
 }
-
 #results {
     height: 200px;
 }
 
-#dice-result{
-    
+input {
+    width: 35px;
+    background-color: paleturquoise;
+}
+#dice-para {
+    display: flex;
+    align-items: center;
+    zoom: 2;
 }
--- a/dice.css
+++ b/dice.css
@@ -1,0 +1,17 @@
+header {
+    display: flex;
+    ;
+}
+section {
+    border: navy 2px solid;
+    padding: 1%;
+    margin: 2.5%  ;
+}
+
+#results {
+    height: 200px;
+}
+
+#dice-result{
+    
+}
--- a/dice.js
+++ b/dice.js
@@ -1,4 +1,3 @@
-
 const diceFaces = ["&#9856", "&#9857", "&#9858", "&#9859", "&#9860", "&#9861"];
 const form = document.querySelector("form");
 const numOfDice = document.querySelector("#num-of-dice");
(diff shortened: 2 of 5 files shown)


dice.css
```diff
@@ -0,0 +1,20 @@
...
+}
+
+input {
+    width: 35px;
+    background-color: paleturquoise;
+}
+#dice-para {
+    display: flex;
+    align-items: center;
+    zoom: 2;
```
Sparkadelic177, line 19: cool use of the css zoom property

dice.js
```diff
@@ -0,0 +1,32 @@
...
+const form = document.querySelector("form");
+const numOfDice = document.querySelector("#num-of-dice");
+const rollButton = document.querySelector("#rollButton");
+let diceResult = document.querySelector("#dice-para");
+let sumText = document.querySelector("#sum-para");
+const originalSumtext = sumText.textContent;
+let ul = document.querySelector("#history-list");
+let dice;
+
+form.button.addEventListener("click", (e) => {
```
Sparkadelic177, line 11: we can also use the 'submit' event listeners on the form

dice.js
```diff
@@ -0,0 +1,32 @@
...
+  diceResult.innerHTML = dice;
+  sumText.textContent += sum;
+  let li = ""
+  li = document.createElement("li");
+  li.innerHTML += dice;
+  li.innerText += ` = ${sum}`;
+  button.addEventListener("click", (e) => { 
+    ul.appendChild(li);
+
+  },);
```
Sparkadelic177, line 30: I see that you have used this event listener to add the history subsequently, after the first click, this event listeners gets defined and now we have 2 event listeners for the same button. To append the history subsequently, we can create a separate function to check to see if there is already something in the `diceResult.innerHTML` before giving it a new value . If so then add the current `diceResult.innerHTML`  in the history then add the new `display.innerHTML` value.

dice.js
```diff
@@ -0,0 +1,32 @@
...
+const originalSumtext = sumText.textContent;
+let ul = document.querySelector("#history-list");
+let dice;
+
+form.button.addEventListener("click", (e) => {
+  e.preventDefault();
+  let sum = 0;
+  sumText.textContent = originalSumtext;
+  dice = "";
+  for (let i = 0; i < Number(numOfDice.value); i++) {
```
Sparkadelic177, line 16: instead of use `numOfDice.valueAsNumber` its slightly cleaner

dice.js
```diff
@@ -0,0 +1,32 @@
...
+    let randomNum = Math.floor(Math.random() * diceFaces.length);
+    dice += diceFaces[randomNum];
+    sum += randomNum + 1;
+  }
+  diceResult.innerHTML = dice;
+  sumText.textContent += sum;
+  let li = ""
+  li = document.createElement("li");
```
Sparkadelic177, line 24: we can add this into one line like so: ` let li = document.createElement("li")`

PR #54, opened 2021-02-28
Carlo Echeverri -  Module Two Mid Assessment

carloecheverri:


index.html
```diff
@@ -0,0 +1,26 @@
...
+    <script src="index.js" defer></script>
+    <title>module 2 test</title>
+</head>
+<body>
+<h1>Dice Roller</h1>
+    Number of dice:  
+    <form>
+    <input type="number" id="dice-input"><br>
+    <button type="submit">Roll the dice</button>
+    </form>
```
Sparkadelic177, line 16: We could of also placed this inside of a section tag

index.js
```diff
@@ -0,0 +1,50 @@
...
+  },
+  {
+    name: "⚄",
+    value: 5,
+  },
+  {
+    name: "⚅",
+    value: 6,
+  },
+];
```
Sparkadelic177, line 26: Nice work, this is awesome

index.js
```diff
@@ -0,0 +1,50 @@
...
+  let sum = 0;
+  const numberInput = document.querySelector("#dice-input");
+  for (let i = 0; i < numberInput.valueAsNumber; i++) {
+    const randomDice = Math.floor(Math.random() * allDice.length);
+    p1.textContent += allDice[randomDice].name;
+    sum += Number(allDice[randomDice].value);
+    p2.textContent = `Sum = ${sum}`;
+  }
+
+  const ul = document.querySelector("ul");
```
Sparkadelic177, line 44: We don't need to grab the `ul` here for every form submission. This can be done just once

index.js
```diff
@@ -0,0 +1,50 @@
...
+  const li = document.createElement("li");
+  li.textContent = `${p1.textContent} = ${sum}`;
+  ul.appendChild(li);
```
Sparkadelic177, line 47: To append the history subsequently, we can create a some logic to check to see if there is already something in the `p1.innerContent` before giving it its value. If so then add the current `p1.innerContent`  in the history then add the new `p1.innerContent` value.

PR #57, opened 2021-02-28
Durdona Djalilova

durdonadjalilova:


index.html
```diff
@@ -0,0 +1,30 @@
...
+    <link rel="stylesheet" href="index.css">
+    <script src = "index.js" defer></script>
+</head>
+<body>
+    <h1>Dice Roller</h1>
+<form id = "dice-form">
+    <label for="number-input">Number of dice: </label>
+    <input type="number" id = "number-input">  <br> <br> 
+    <button type = "submit">Roll the dice!</button> 
+</form> <br>
```
Sparkadelic177, line 17: This can be placed inside of another section tag, and instead of using the `br` tag, try using flexbox for positioning

index.js
```diff
@@ -0,0 +1,63 @@
...
+    const input = document.querySelector("#number-input")
+    const button = document.querySelector("button")
+    const outputDice = document.querySelector("#dice-para")
+    let outputSum = document.querySelector("#sum-para")
+    const ul = document.querySelector("#history-list")
+    const formDice = document.querySelector("dice-form") 
+
+    button.addEventListener("click", (e) => {
+       addToHistory(e)
+        rollTheDice(e)
```
Sparkadelic177, line 19: we can call `e.preventDefault()` here inside of the callback function instead of doing it in both function

Sparkadelic177, line 19: No need to call `rollTheDice` here since you are calling this function inside of the `addToHistory` function

index.js
```diff
@@ -0,0 +1,63 @@
...
+    })
+   
+    let count = 0
+    
+    
+    function rollTheDice (e) {
+        e.preventDefault()
+            let sum = 0
+            for(let i =0; i < Number(input.value); i++) {
+         let randomDice = dice[Object.keys(dice)[Math.floor(Math.random()*Object.keys(dice).length)]]//dice[Math.floor(Math.random() * dice.length)]
```
Sparkadelic177, line 29: try to not repeat yourself here to many times

index.js
```diff
@@ -0,0 +1,63 @@
+
+const dice = {"&#9856" : "⚀",
+    "&#9857" : "⚁",
+    "&#9858" : "⚂",
+    "&#9859" :"⚃",
+    "&#9860" :"⚄",
+    "&#9861" : "⚅"}
```
Sparkadelic177, line 7: We can also make an object where we have `dice:"&#9857"` and `value:2`. Think about how the rest of your code would change

PR #58, opened 2021-02-28
Farrah Rios

rios-isabelle-farrah:
Thank you for checking in on me!

### DIFF
--- a/index.html
+++ b/index.html
@@ -20,8 +20,8 @@
   
     </section>
 
-    <section> 
-        <ul id='history-dice'></ul>
+    <section id ='second'>
+   <ul></ul>
      
     </section>
    
--- a/index.html
+++ b/index.html
@@ -1,0 +1,40 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <script src="index.js" defer></script>
+    <link rel="stylesheet" href="style.css">
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Document</title>
+</head>
+<body>
+    <h1></h1>
+    <section id="first">  
+<form action="">
+    <input type="number">
+<button>Roll Dice</button>
+<p id='dice-para'></p>
+<p id='sum-para'></p>
+</form>
+  
+    </section>
+
+    <section> 
+        <ul id='history-dice'></ul>
+     
+    </section>
+   
+
+
+
+    <!-- <p>
+&#9856 = ⚀
+&#9857 = ⚁
+&#9858 = ⚂
+&#9859 = ⚃
+&#9860 = ⚄
+&#9861 = ⚅
+    </p>  -->
+</body>
+</html>
--- a/index.js
+++ b/index.js
@@ -1,113 +1,76 @@
+let oneD = "⚀";
+let twoD = "⚁";
+let threeD = "⚂";
+let fourD = "⚃";
+let fiveD = "⚄";
+let sixD = "⚅";
+let conditional = 0;
+let allItems = document.createElement("ul");
+allItems.id = "history-dice";
 
+let arrayOfDice = [oneD, twoD, threeD, fourD, fiveD, sixD];
+let arrayOfDiceRolls = [1, 2, 3, 4, 5, 6];
 
-
-
-let oneD = "⚀" 
-let twoD = '⚁'
-let threeD = '⚂'
-let fourD = '⚃'
-let fiveD = '⚄'
-let sixD = '⚅'
-let conditional = true
-
-
-let arrayOfDice = [oneD,twoD,threeD,fourD,fiveD,sixD]
-let arrayOfDiceRolls = [1,2,3,4,5,6]
-
-let button = document.querySelector('button');
-button.addEventListener('click', addDice);
+let button = document.querySelector("button");
+button.addEventListener("click", addDice);
 
 function addDice(e) {
+  e.preventDefault();
 
-    e.preventDefault();
- 
-    let numberInput = document.querySelector("input")
-    let stringOfOutput = ""
-    let sum = 0
-    let historyDiceList = document.querySelector('#history-dice')
-    let newLi = document.createElement('li')
-    newLi.id = 'HistoryListItem'
-for (let i=1;i<= numberInput.value;i++){
- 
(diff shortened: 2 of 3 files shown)


index.html
```diff
@@ -0,0 +1,40 @@
...
+    <script src="index.js" defer></script>
+    <link rel="stylesheet" href="style.css">
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Document</title>
+</head>
+<body>
+    <h1></h1>
+    <section id="first">
```
coreyladovsky, line 13: Rethink id names to be more descriptive

index.html
```diff
@@ -0,0 +1,40 @@
...
+    <h1></h1>
+    <section id="first">  
+<form action="">
+    <input type="number">
+<button>Roll Dice</button>
+<p id='dice-para'></p>
+<p id='sum-para'></p>
+</form>
+  
+    </section>
```
coreyladovsky, line 21: Nice semantic tags

index.html
```diff
@@ -0,0 +1,40 @@
...
+<head>
+    <script src="index.js" defer></script>
+    <link rel="stylesheet" href="style.css">
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Document</title>
+</head>
+<body>
+    <h1></h1>
```
coreyladovsky, line 12: Why is your h1 empty?

index.html
```diff
@@ -0,0 +1,40 @@
...
+<form action="">
+    <input type="number">
+<button>Roll Dice</button>
+<p id='dice-para'></p>
+<p id='sum-para'></p>
+</form>
+  
+    </section>
+
+    <section id ='second'>
```
coreyladovsky, line 23: Remember to format all of your code. Use a formatter like prettier to help you. Option + shift + f

index.html
```diff
@@ -0,0 +1,40 @@
...
+
+
+
+    <!-- <p>
+&#9856 = ⚀
+&#9857 = ⚁
+&#9858 = ⚂
+&#9859 = ⚃
+&#9860 = ⚄
+&#9861 = ⚅
```
coreyladovsky, line 37: Remember to removed dead code :-)

index.js
```diff
@@ -0,0 +1,76 @@
+let oneD = "⚀";
+let twoD = "⚁";
+let threeD = "⚂";
+let fourD = "⚃";
+let fiveD = "⚄";
+let sixD = "⚅";
```
coreyladovsky, line 6: could be const b/c not reassigned

Also we don't need these. Just make them the values in arrayOfDice.

index.js
```diff
@@ -0,0 +1,76 @@
...
+let conditional = 0;
```
coreyladovsky, line 7: conditional sounds like a boolean, seeing it equal 0 is confusing

index.js
```diff
@@ -0,0 +1,76 @@
...
+let allItems = document.createElement("ul");
```
coreyladovsky, line 8: I want you to start think about when you should be using const apposed to let. let seems to be your default, switch that behavior.

index.js
```diff
@@ -0,0 +1,76 @@
...
+allItems.id = "history-dice";
+
+let arrayOfDice = [oneD, twoD, threeD, fourD, fiveD, sixD];
+let arrayOfDiceRolls = [1, 2, 3, 4, 5, 6];
+
+let button = document.querySelector("button");
+button.addEventListener("click", addDice);
```
coreyladovsky, line 15: should listen for "submit" on form instead of button "click".

index.js
```diff
@@ -0,0 +1,76 @@
...
+  e.preventDefault();
+
+  let newLi = document.createElement("li");
+  newLi.id = "completeList";
+
+  let numberInput = document.querySelector("input");
+  let stringOfOutput = "";
+  let sum = 0;
+
+  for (let i = 1; i <= numberInput.value; i++) {
```
coreyladovsky, line 27: why not set i = 0 and do < numberInput.valueAsNumber ? instead of <=

index.js
```diff
@@ -0,0 +1,76 @@
...
+    switch (
```
coreyladovsky, line 28: Cool that you're using switch. 
But much of this can be simplified by getting the index, and then indexing into your array and recognizing the die value is the index + 1

index.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+  let currentDiceOutput = document.querySelector("#dice-para");
+  currentDiceOutput.textContent = stringOfOutput;
+  let sumOfDiceOutput = document.querySelector("#sum-para");
+  updateOutput = currentDiceOutput.textContent;
+  sumOfDiceOutput.textContent = `Sum of Die: ${sum}`;
+  conditional += 1;
+  // debugger
+
+  newLi.textContent = currentDiceOutput.textContent;
```
coreyladovsky, line 67: I'd create this right before I use it so it's easier for a reader to keep track of everything

index.js
```diff
@@ -0,0 +1,76 @@
...
+      // code block
+    }
+  }
+  let currentDiceOutput = document.querySelector("#dice-para");
+  currentDiceOutput.textContent = stringOfOutput;
+  let sumOfDiceOutput = document.querySelector("#sum-para");
+  updateOutput = currentDiceOutput.textContent;
+  sumOfDiceOutput.textContent = `Sum of Die: ${sum}`;
+  conditional += 1;
+  // debugger
```
coreyladovsky, line 65: Delete dead code

index.js
```diff
@@ -0,0 +1,76 @@
...
+
+  newLi.textContent = currentDiceOutput.textContent;
+
+  allItems.appendChild(newLi);
```
coreyladovsky, line 69: We don't want to always append this right away. We want to append the old value that was there if it exists. Then we don't have to worry about when the ul should be showing up. We just need to change our order of when we do things.

style.css
```diff
@@ -0,0 +1,19 @@
...
+   padding: 0px;
+   border:0ch;
+   margin:0px;
+}
+input{
+    height: 40px;
+    width: 80px;
+    font-size: xx-large;
+}
+p#sum-para{
```
coreyladovsky, line 12: ids are unique. So p#sum-para shouldn't be used. just #sum-para

PR #59, opened 2021-02-28
assesment complete

Teraesa:


index.html
```diff
@@ -0,0 +1,47 @@
...
+    <link rel="stylesheet" href="index.css"/>
+    <title>Dice Roller</title>
+</head>
+<body>
+    <h1>Dice Roller</h1>
+
+
+   <section>
+    <form>
+        <label>Number of dice:</label>
```
coreyladovsky, line 16: labels should either wrap inputs or point to the input it's relating to. Needs the for attribute and points to input id.

index.html
```diff
@@ -0,0 +1,47 @@
...
+        <input id="dice-input" type="number"/>
+        <br>
```
coreyladovsky, line 18: Be wary of using line break tags. Best practice is control position of elements via css

index.html
```diff
@@ -0,0 +1,47 @@
...
+
+
+
+        <button type="number" id="roll-dice-btn">Roll the dice</button>
+    </form>
+    
+</section>
+
+<div id="dice-sum">
```
coreyladovsky, line 27: use semantic html prefer section to div

index.html
```diff
@@ -0,0 +1,47 @@
...
+
+    <span>
+       <p id="display-image"> &#9856;
+        &#9857;
+       &#9858;
+       &#9859;
+       &#9860;
+       &#9861;
```
coreyladovsky, line 35: These die are here permanently. We only want to show the die that were rolled.

index.js
```diff
@@ -0,0 +1,86 @@
+const dice = [
+  "&#9856:1",
```
coreyladovsky, line 2: This is definitely the right though. I don't think we need the :number b/c the number is their index + 1

index.js
```diff
@@ -0,0 +1,86 @@
...
+const diceButton = document.querySelector("#roll-dice-btn"); //button to submit
+
+const h2 = document.querySelector("#sum-para"); //this prints history to screen
+const ul = document.querySelector("ul");
+
+const p = document.querySelector("#display-image");
+let original = "Number of dice:";
+//let count = 0; //or 0?
+let diceList = ""; //number of times dice is rolled
+let sumOfInput = ""; // hold inputs from numinput
```
coreyladovsky, line 24: sum sounds like a number so should probably point to 0 instead of ""

index.js
```diff
@@ -0,0 +1,86 @@
...
+let sum;
+
+//Functions
+
+//get dice to print
+const getDice = (event) => {
+  event.preventDefault();
+
+  //pick the amount  needed numinPut.value * , also need a loop
+  for (let i = 0; i < Number(numInput.value); i++) {
```
coreyladovsky, line 38: .valueAsNumber for slightly cleaner code

index.js
```diff
@@ -0,0 +1,86 @@
...
+    let randomRoll = Math.floor(Math.random() * 6) + 1;
+    let total = randomRoll + numInput;
```
coreyladovsky, line 40: numInput here is currently and entire htmlElement. We don't want to do math with it. 
By declaring total inside of the for loop, you're getting a new variable with each iteration. this does not update the total variable previously declared.

index.js
```diff
@@ -0,0 +1,86 @@
+const dice = [
+  "&#9856:1",
+  "&#9857:2",
+  "&#9858:3",
+  "&#9859:4",
+  "&#9860:5",
+  "&#9861:6",
+];
```
coreyladovsky, line 8: what about using the die icons instead? ["⚀",  "⚁", ....etc] improve your code

index.js
```diff
@@ -0,0 +1,86 @@
...
+    //get dice and store in diceList variable
+    diceList += randomRoll;
+
+    //display dice and reset input area
+
+    diceOutput.textContent = `Number of dice: ${diceList}`;
+    //numInput.value = "";
+  }
+};
+diceButton.addEventListener("click", getDice);
```
coreyladovsky, line 50: should listen for "submit" on form instead of button "click".

index.js
```diff
@@ -0,0 +1,86 @@
...
+    listItem.textContent = `${diceList}= ${total}`; //dicelist (dices) and number total
+    //diceList = "";
+
+    diceOutput = "History";
+    numInput.value = "";
+    //count++;
+  });
+};
+
+diceButton.addEventListener("click", diceHistory);
```
coreyladovsky, line 70: should listen for "submit" on form instead of button "click". 

The preventDefault is b/c we want to stop the refresh behavior of a form submission.

index.js
```diff
@@ -0,0 +1,86 @@
...
+    diceOutput.textContent = `Number of dice: ${diceList}`;
+    //numInput.value = "";
+  }
+};
+diceButton.addEventListener("click", getDice);
+
+//add history
+const diceHistory = (event) => {
+  event.preventDefault();
+  let newArray = numInput.value.split("\n"); //numinput is the accumulater,b is the current value
```
coreyladovsky, line 55: numInput.value is going to give a number as a string. We don't want to split it on new lines, b/c the number won't have any new lines. 

This should result in an array with one element in it, that element being the same initial number as a string.

index.js
```diff
@@ -0,0 +1,86 @@
...
+  newArray.map((el) => {
+    listItem = document.createElement("li");
```
coreyladovsky, line 57: We want to create the new li, after each roll. So only one at a time. No need for iterating here.

index.js
```diff
@@ -0,0 +1,86 @@
...
+
+const p = document.querySelector("#display-image");
+let original = "Number of dice:";
+//let count = 0; //or 0?
+let diceList = ""; //number of times dice is rolled
+let sumOfInput = ""; // hold inputs from numinput
+let listItem; //to hold li
+let currentValue = 0; //counter
+let num;
+let total;
```
coreyladovsky, line 28: total should probably have an initial assignment if you plan to use it for math. Currently has the value of undefined.

index.js
```diff
@@ -0,0 +1,86 @@
...
+    diceOutput = "History";
+    numInput.value = "";
+    //count++;
+  });
+};
+
+diceButton.addEventListener("click", diceHistory);
+
+//adds the sum printed when inputs are added
+const sumOfOutput = (event) => {
```
coreyladovsky, line 73: Instead of having a different function here, we should be adding the sum of the die, as we're rolling them in your getDice function.

index.js
```diff
@@ -0,0 +1,86 @@
...
+  sum += sumPrinted;
+  total++;
+  if (total < 0) {
+    h2.textContent = numInput + sum + total;
+  } else {
+    h2.textContent = "";
+  }
+};
+
+diceButton.addEventListener("click", sumOfOutput);
```
coreyladovsky, line 86: We don't want have 3 listeners all listening for the exact same event. Instead have one listener. The callback function for that can call other functions.

index.js
```diff
@@ -0,0 +1,86 @@
...
+  });
+};
+
+diceButton.addEventListener("click", diceHistory);
+
+//adds the sum printed when inputs are added
+const sumOfOutput = (event) => {
+  event.preventDefault();
+  let currentValue = numInput.value + diceList;
+  let sumPrinted = numInput.value + currentValue; // accumlator = numinput , currentvalue to be added to accumlator
```
coreyladovsky, line 76: I'm seeing a lot of reference to an accumulator but I'm not seeing any of that in your code...

PR #60, opened 2021-02-28
Dalíz

dalizcruz:


index.js
```diff
@@ -0,0 +1,26 @@
...
+
+const savedHistory = (event) => {
+   event.preventDefault();
+  const ul = document.querySelector("ul");
+  const li = document.createElement("li");
+  li.textContent = dieResult.textContent;
+  ul.appendChild(li);
+};
+form.addEventListener("submit", rollDice);
+form.addEventListener("submit", savedHistory);
```
Sparkadelic177, line 26: its not recommended to have two event listeners for the same element. We can call the savedHistory function with in the rollDice function

index.js
```diff
@@ -0,0 +1,26 @@
+const form = document.querySelector("#num-of-dice-form"); 
+const dieResult = document.querySelector("#dice-para"); 
+
+const rollDice = (event) => {
+  event.preventDefault();
+  dieResult.textContent = "";
+  const numValue = document.querySelector("#num-input"); 
+  for (let i = 0; i < numValue.value; i++) {
```
Sparkadelic177, line 8: be aware of the data types that you're using `numValue.value` returns a string we can also use `.valueAsNumber` to return a number data type

index.js
```diff
@@ -0,0 +1,26 @@
...
+
+  
+};
+
+const savedHistory = (event) => {
+   event.preventDefault();
+  const ul = document.querySelector("ul");
+  const li = document.createElement("li");
+  li.textContent = dieResult.textContent;
+  ul.appendChild(li);
```
Sparkadelic177, line 23: To append the history subsequently, we can create a separate function to check to see if there is already something in the `dieResult.textContent` before giving it its value. If so then add the current `dieResult.textContent`  in the history then add the new `dieResult.textContent` value.

PR #61, opened 2021-02-28
Kalilah Clarke

KalilahClarke:
Part One

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,32 @@
+const die= { "⚀": 1, "⚁": 2, "⚂": 3, "⚃": 4, "⚄": 5, "⚅": 6 }
+const keys = Object.keys(die)
+
+document.addEventListener("DOMContentLoaded", () => {
+    const form = document.querySelector("form")
+    const ul = document.querySelector("ul");
+    const p = document.querySelector("p");
+    const roll = document.querySelector("#roll-it")
+    
+    form.addEventListener("submit",(event) => {
+        // html form connects to addEventListen("submit")
+        event.preventDefault();
+        // stop website from refreshing look at the refresh button and the url to see if it is actually working (?) means preventDefault isn't being called
+        const number = Number(document.querySelector("#number-dice").value);
+        let string = "";
+        if (number === 0){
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -0,0 +1,32 @@
...
+    const form = document.querySelector("form")
+    const ul = document.querySelector("ul");
+    const p = document.querySelector("p");
+    const roll = document.querySelector("#roll-it")
+    
+    form.addEventListener("submit",(event) => {
+        // html form connects to addEventListen("submit")
+        event.preventDefault();
+        // stop website from refreshing look at the refresh button and the url to see if it is actually working (?) means preventDefault isn't being called
+        const number = Number(document.querySelector("#number-dice").value);
```
Sparkadelic177, line 14: Instead of using `.value` and using the `Number()` function we can use .`valueAsNumber`, its slightly cleaner

index.js
```diff
@@ -0,0 +1,32 @@
...
+        let string = "";
+        if (number === 0){
+            p.textContent = "Error: Please choose a Number > 0"
+        }else{
+            for(let i = 0; i < number; i++){
+                let random = Math.floor(Math.random() * Object.keys(die).length);
```
Sparkadelic177, line 20: we could of also used 'keys.length` instead of repeating ourself with `Object.keys(die).length`

index.js
```diff
@@ -0,0 +1,32 @@
...
+                let dices = Object.keys(die)[random]; 
+                string += dices
+                
+            }
+            p.textContent = `${string}`
+         }  
+        //What's missing is the sum which is the object.value that needs to be keyed into and sumed up within a function
+        // then all information needed to be summed up and storied in history with the use of the document.createElement("li");
+        // then needs to be appended to the parent directory ul
```
Sparkadelic177, line 29: I like these notes that you have here for yourself, the sum could've  been calculated with the index of the keys array plus 1. For example `keys[random] / keys[0]` is the die with the value of 1, and in the `for loop` we can get the sum like so: `sum += random + 1`

PR #62, opened 2021-02-28
Flower Aviles

floweraviles:


index.js
```diff
@@ -0,0 +1,35 @@
...
+const dicePara = document.querySelector("#dice-para")
+const sumPara = document.querySelector("#sum-para")
+const historyTitle = document.querySelector("#history-title")
+
+
+const diceCount = (e) => {
+    e.preventDefault();
+    dicePara.textContent = "";
+    sumPara.textContent = "Sum = ";
+    sum = 0
```
Sparkadelic177, line 19: don't forget to add the define this variable

index.js
```diff
@@ -0,0 +1,35 @@
...
+    const numberInput = document.querySelector("#dice-length-input")
+    for (let i = 0; i < numberInput.valueAsNumber; i++) {
```
Sparkadelic177, line 21: I love the use of the valueAsNumber property

index.js
```diff
@@ -0,0 +1,35 @@
...
+        const randIdx = Math.floor(Math.random() * dice.length)
+        dicePara.textContent += dice[randIdx]
+        sum += randIdx+ 1
+        
+    }
+    sumPara.textContent = "Sum = " + sum
+    historyTitle.textContent = "History"
+    const ul = document.querySelector("ul");
```
Sparkadelic177, line 29: We don't need to grab the ul here for every form submission. This can be done just once

index.js
```diff
@@ -0,0 +1,35 @@
...
+    const li = document.createElement("li");
+    li.textContent = `${dicePara.textContent} = ${sum}`
+    ul.appendChild(li);
```
Sparkadelic177, line 32: To append the history subsequently, we can create a separate function to check to see when if there is already something in the `display.innerText`. If so then add the current `display.innerText`  in the history then add the new display.innerText value.

PR #63, opened 2021-02-28
Demi Mid Mod

demi5456:


dice.js
```diff
@@ -0,0 +1,58 @@
+// const dice = ['&#9856 = ⚀', '&#9857 = ⚁', '&#9858 = ⚂', '&#9859 = ⚃', '&#9860 = ⚄', '&#9861 = ⚅']
+let dice = [{ 1: '⚀', 2: '⚁', 3: '⚂', 4: '⚃', 5: '⚄', 6: '⚅' }]
```
coreyladovsky, line 2: I like some of this, but I'm not sure why we're putting this object in an array? 
If we want an array could just do [ '⚀',  '⚁', ... etc ] then index + 1 is value

dice.js
```diff
@@ -0,0 +1,58 @@
...
+let randomNum = dice[Math.floor(Math.random() * dice.length)]
```
coreyladovsky, line 3: could be const b/c not reassigned

coreyladovsky, line 3: dice length is always 1 and dice at 0 is the object

dice.js
```diff
@@ -0,0 +1,58 @@
...
+
+
```
coreyladovsky, line 5: It appears your default variable declaration is let. Switch that to const

dice.js
```diff
@@ -0,0 +1,58 @@
...
+let rollBtn = document.querySelector("#roll-btn")
+let numInput = document.querySelector("#number")
+let diceHolder = document.querySelector("#dice-para")
```
coreyladovsky, line 8: Nice variable names, keeps it easy to read

dice.js
```diff
@@ -0,0 +1,58 @@
...
+let display = ""
+let display2 = ""
+let click = 0
+let sum1 = 0
+let sum2 = 0
+
+
+
+const diceRoll = () => {
+    for (let i = 1; i <= numInput.value; i++) {
```
coreyladovsky, line 21: .valueAsNumber for slightly cleaner code
start i at 0 and while < instead of 1 and <=

dice.js
coreyladovsky: what's this line for? Seems like we get a random number later in diceRoll

dice.js
```diff
@@ -0,0 +1,58 @@
...
+        let randDice = Math.floor(Math.random() * 6) + 1 //generate a new random # btwn 1 & 6 per iteration
+        display += randomNum[randDice]
+        sum1 += randDice
```
coreyladovsky, line 24: I like this

dice.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+    diceHolder.textContent = display //display dice as string
+    sumHolder.textContent = `Sum = ${sum1}`
+    sum2 = sum1
+    display2 = display
+}
+
+const saveRoll = () => {
+    console.log(click)
```
coreyladovsky, line 33: remove console logs and or any dead code

dice.js
```diff
@@ -0,0 +1,58 @@
...
+        display += randomNum[randDice]
+        sum1 += randDice
+    }
+    diceHolder.textContent = display //display dice as string
+    sumHolder.textContent = `Sum = ${sum1}`
+    sum2 = sum1
+    display2 = display
+}
+
+const saveRoll = () => {
```
coreyladovsky, line 32: I like that you tried breaking this into it's own function. It's still got some duplicate code in there that could be broken down further. 

I think the switching of the sum1 and sum2 is a bit confusing for me. Why not update the list before refilling the value in the para?

dice.js
```diff
@@ -0,0 +1,58 @@
...
+        list.textContent = `${display2} = ${sum2}`
+        ul.appendChild(list)
+    }
+    click++
+}
+
+
+
+
+rollBtn.addEventListener("click", e => {
```
coreyladovsky, line 52: should listen for "submit" on form instead of button "click".

index.html
```diff
@@ -0,0 +1,26 @@
...
+    <title>diceRolling</title>
+</head>
+<body>
+<h1>Dice Roller</h1>
+<form>
+ Number of dice:   <input type="number" id="number">
+ <br>
+ <button id="roll-btn">Roll the dice!</button>
+</form>
+<div id="showDice">
```
coreyladovsky, line 17: use semantic html prefer section to div

index.html
```diff
@@ -0,0 +1,26 @@
...
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <script src="./dice.js" defer></script>
+    <title>diceRolling</title>
+</head>
+<body>
+<h1>Dice Roller</h1>
+<form>
+ Number of dice:   <input type="number" id="number">
+ <br>
```
coreyladovsky, line 14: Be wary of using line break tags. Best practice is control position of elements via css

index.html
```diff
@@ -0,0 +1,26 @@
...
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <script src="./dice.js" defer></script>
+    <title>diceRolling</title>
+</head>
+<body>
+<h1>Dice Roller</h1>
+<form>
+ Number of dice:   <input type="number" id="number">
```
coreyladovsky, line 13: Use a label here and either wrap it around the input or point it at the input id with the label `for` attribute

index.html
```diff
@@ -0,0 +1,26 @@
...
+<div id="showDice">
+<p id="dice-para"></p>
+<p id="sum-para"></p>
+</div>
+<div id="history">
+    <h2>History</h2>
+    <ul></ul>
+</div>
+</body>
+</html>
```
coreyladovsky, line 26: Remember to use a code formatter like prettier on all files. keyboard shortcut option + shift + f

PR #65, opened 2021-03-02
Assessment attempt

myrasmith:


dice.js
```diff
@@ -0,0 +1,9 @@
+document.addEventListener('DOMContentLoaded', () => {
+    const btn = document.querySelector('button')
+    btn.addEventListener('click', () => {
+        console.log('clicked!')
+        const input = document.querySelector('input')
+        console.log(input)
```
myrasmith, line 6: Is the input variable a number?

